PR #4, opened 2017-07-28
Fix realm, broker, dealer close

gammazero:


router/broker.go
```diff
@@ -117,28 +115,13 @@ func (b *broker) RemoveSession(sub *Session) {
 
 // Close stops the broker and waits message processing to stop.
 func (b *broker) Close() {
-	if b.Closed() {
-		return
-	}
 	close(b.reqChan)
```
mourad: There are various functions sending messages over `reqChan`, such as, `Submit` and `RemoveSession` that would cause a `panic`.

gammazero: Thought I guarantees they would all be done by the time this was closed.  Will re-evaluate.

PR #6, opened 2017-07-28
Regoncreate

gammazero:
Complete registration meta events:
wamp.registration.on_create
wamp.registration.on_register
wamp.registration.on_unregister
wamp.registration.on_delete

### DIFF
--- a/router/dealer.go
+++ b/router/dealer.go
@@ -343,7 +343,10 @@ func (d *dealer) register(callee *Session, msg *wamp.Register) {
 	})
 
 	// Publish wamp.registration.on_register meta event.  Fired when a session
-	// is added to a registration.
+	// is added to a registration.  A wamp.registration.on_register event MUST
+	// be fired subsequent to a wamp.registration.on_create event, since the
+	// first registration results in both the creation of the registration and
+	// the addition of a session.
 	d.metaClient.Send(&wamp.Publish{
 		Request:   wamp.GlobalID(),
 		Topic:     wamp.MetaEventRegOnRegister,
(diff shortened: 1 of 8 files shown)


mourad:
The 3 commits related with meta api look ok to me, can you rebase to remove the commits related with `close`?

Also, you'll need to rebase on top of `master` to resolve the conflict with `realm.go`

router/dealer.go
```diff
@@ -160,8 +160,11 @@ func NewDealer(strictURI, allowDisclose bool, metaClient wamp.Peer) Dealer {
 // Submit dispatches a Resister, Unregister, Call, Cancel, Yield, or Error
 // message to the dealer.
 func (d *dealer) Submit(sess *Session, msg wamp.Message) {
-	if msg == nil || sess == nil {
-		panic("nil session or message")
+	if msg == nil {
+		panic("dealer.Submit with nil message")
+	}
+	if sess == nil {
+		panic("dealer.Submit with nil session")
```
mourad: Seems wrong for the dealer to panic.

router/dealer.go
```diff
@@ -218,6 +220,9 @@ func (d *dealer) reqHandler() {
 			case RegCountCallees:
 				rsp = d.regCountCallees(msg)
 			default:
+				// This is a programming error because, a client cannot submit
+				// an INVACATION to the dealer, only the realm's meta procedure
```
mourad: INVOCATION

router/dealer.go
```diff
@@ -597,7 +602,7 @@ func (d *dealer) call(caller *Session, msg *wamp.Call) {
 		Arguments:    msg.Arguments,
 		ArgumentsKw:  msg.ArgumentsKw,
 	})
-	//log.Printf("dispatched CALL: %v [%v] to callee as INVOCATION %v",
+	//log.Printf("Dispatched CALL: %v [%v] to callee as INVOCATION %v",
```
mourad: should this comment be removed?

router/dealer.go
gammazero: panics like this are reserved for places that could ONLY be caused by programming errors.  If this were not an exported interface, I could remove it, but it means illegal misuse of the interface, therefore panic.

router/dealer.go
gammazero: Yes, and a few other like it.  I left them in there in a commented out state, since I thought I may use them for debugging, but really, they should be removed. Will do.

PR #7, opened 2017-07-29
Use context.Context for call cancellation, timeout, and cancellation …

gammazero:
…notification.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -875,7 +875,7 @@ func (c *Client) waitForReplyWithCancel(ctx context.Context, id wamp.ID, mode st
 	select {
 	case msg = <-wait:
 	case <-ctx.Done():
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -862,11 +862,20 @@ func (c *Client) waitForReplyWithCancel(id wamp.ID, cancelAfter time.Duration, m
...
+		defer cancel()
+	}
+
 	var msg wamp.Message
 	var err error
 	select {
 	case msg = <-wait:
-	case <-time.After(cancelAfter):
+	case <-ctx.Done():
+		log.Print("Caller cancelling call %v (mode=%s)", id, mode)
```
mourad: This log entry seems a bit confusing. Isn't it possible that the call was cancelled due to the deadline (and nothing the caller did)?

gammazero: good point, rewording required

PR #9, opened 2017-07-30
Fixing close

mourad:


mourad:
I can definitely see the benefit of preventing locks during the normal execution path.

PR #10, opened 2017-07-30
Guarantee safe close - alternate locking strategy

gammazero:
Closing realm is guaranteed safe by the following, plus changes in this PR:

- Realm guarantees there are never multiple calls to `broker.Close()` or `dealer.Close()`
- `handleSession()` and `metaProcedureHandler()` are the only things than can call broker/dealer.`Submit()` or broker/dealer.`RemoveSession()`
- Closing Realm waits until all `realm.handleSession()` and `realm.metaProcedureHandler()` have exited, thereby guaranteeing that there will be no more `broker/dealer.Submit()` or `broker/dealer.RemoveSession()`, therefore no chance of submitting to closed channel.
- Finally realm closes broker and dealer `reqChan`, which is safe.  Even if broker or dealer are in the process of publishing meta events or calling meta procedures, there is no `metaProcedureHandler()` to call `Submit()`.  Any client sessions that are still active likewise have no `handleSession()` to call `Submit()` or `RemoveSession()`.

The safety exception exception is if the router was attaching a new client and execution got to here (https://github.com/gammazero/nexus/blob/master/router/router.go#L291), but had not yet incremented the counter, and all the close logic had completed just before that, then a new `handleSession()` could be started on a closed realm.

This change prevents the exception by holding a lock while checking if the realm is closed and incrementing the wait counter.  The lock is held in mutual exclusion with the realm closing.

### DIFF
--- a/router/broker.go
+++ b/router/broker.go
@@ -43,6 +43,11 @@ type Broker interface {
 	Features() map[string]interface{}
 }
 
+type brokerReq struct {
+	session *Session
+	msg     wamp.Message
+}
+
 type broker struct {
 	// topic URI -> {subscription ID -> subscribed Session}
 	topicSubscribers    map[wamp.URI]map[wamp.ID]*Session
@@ -57,7 +62,7 @@ type broker struct {
 	// Session -> subscription ID set
 	sessionSubIDSet map[*Session]map[wamp.ID]struct{}
 
-	reqChan chan routerReq
+	reqChan chan brokerReq
 
 	// Generate subscription IDs.
 	idGen *wamp.IDGen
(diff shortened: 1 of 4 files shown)


mourad:
I agree that making sure the realm handles closing the dealer/router at with the correct logic prevents the overhead of locking with every message as discussed in PR#9.

However, the realm code is becoming a bit difficult to reason.  For example, there appears to be at least one deadlock with regards to closing it, if I'm not mistaken. If a new session is coming and is in `onJoin` and has not yet added itself to the `r.clients` map, it is basically here:
https://github.com/gammazero/nexus/blob/altclosefix/router/realm.go#L251
And the realm is being closed here:
https://github.com/gammazero/nexus/blob/altclosefix/router/realm.go#L178

This new session will have already added itself to the `r.waitRealms` WaitGroup but will never receive the close message, so the realm will block indefinitely here:
https://github.com/gammazero/nexus/blob/altclosefix/router/realm.go#L183

With that said, this situation appears to have existed even prior to this PR. I suggest merging this PR anyway, but we should spent some time simplifying both, the realm and router.

gammazero:
I agree that it is very difficult to reason about synchronization during shutdown, and it we must simplify this process if possible.  I will check this is with a note about investigating whether or not a new session, starting while realm is closing can miss its shutdown signal, and make realm wait forever for a session that does not shutdown.

PR #14, opened 2017-08-02
Created NewLocalClient and removed LocalClient from router's interface

mourad:
This PR moves the LocalClient from the router's interface to `client` package. It also moves the `LinkedPeers` from `router/session.go` to `transport/localpeer.go` (following the same pattern as `transport/websocketpeer.go`)

### DIFF
--- a/router/realm.go
+++ b/router/realm.go
@@ -204,9 +204,7 @@ func (r *realm) Run() {
 func (r *realm) createMetaSession() (*Session, *Session) {
 	cli, rtr := transport.LinkedPeers(0, log)
 
-	details := map[string]interface{}{
-		"authrole": "trusted",
-	}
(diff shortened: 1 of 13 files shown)


router/realm.go
```diff
@@ -198,20 +197,16 @@ func (r *realm) Run() {
...
-	if details == nil {
-		details = map[string]interface{}{}
-	} else {
-		wamp.NormalizeDict(details)
+// This is used for creating a local client for publishing meta events.
+func (r *realm) createMetaSession() (*Session, *Session) {
+	cli, rtr := transport.LinkedPeers(0, log)
+
+	details := map[string]interface{}{
+		"authrole": "trusted",
```
gammazero: FYI, this is the same as: `wamp.SetOption(nil, "authrole", "trusted")`

PR #15, opened 2017-08-03
Adding RouterConfig and RealmConfig

mourad:
Also, removed the `realm` interface to keep it internal to the `router` package.


aat/main_test.go
```diff
@@ -51,12 +46,41 @@ func TestMain(m *testing.M) {
...
-	nxr.AddRealm(wamp.URI(testRealm), anonAuth, allowDisclose)
+	routerConfig := &router.RouterConfig{
+		RealmConfigs: []*router.RealmConfig{
+			&router.RealmConfig{
+				URI:           wamp.URI(testRealm),
+				StrictURI:     false,
+				AnonymousAuth: true,
+				AllowDisclose: false,
+				Broker:        true,
+				Dealer:        true,
```
gammazero: I think broker and dealer should be enabled by default, if we even want to give an option to disable them (since the only overhead is a an idle goroutine and small struct).  If we even have the options then should change to: `DisableBroker` and `DisableDealer`  or something similar (`NoBroker`, `BrokerOff`, etc)

aat/main_test.go
```diff
@@ -51,12 +46,41 @@ func TestMain(m *testing.M) {
...
+
 	// Create router instance.
-	nxr = router.NewRouter(autoRealm, strictURI)
-	nxr.AddRealm(wamp.URI(testRealm), anonAuth, allowDisclose)
+	routerConfig := &router.RouterConfig{
+		RealmConfigs: []*router.RealmConfig{
+			&router.RealmConfig{
+				URI:           wamp.URI(testRealm),
+				StrictURI:     false,
+				AnonymousAuth: true,
```
gammazero: How do most apps you know of use WAMP authentication?  Is anonymous the usual, or the exception?  If anonymous in normally the case, then maybe best to default it to allowed.

router/router_config.go
```diff
@@ -0,0 +1,6 @@
+package router
+
+type RouterConfig struct {
+	RealmConfigs  []*RealmConfig `json:"realms"`
+	RealmTemplate *RealmConfig   `json:"realm_template"`
+}
```
gammazero: I would just put this in router.go.

router/realm.go
```diff
@@ -97,33 +107,10 @@ func NewRealm(uri wamp.URI, strictURI, anonymousAuth, allowDisclose bool) Realm
...
-
-// DelAuthenticator deletes the Authenticator for the specified method.
-func (r *realm) DelAuthenticator(method string) {
-	r.actionChan <- func() {
-		delete(r.authenticators, method)
-	}
-	log.Printf("Deleted authenticator for method %s (realm=%v)", method, r.uri)
-}
-
-func (r *realm) SetAuthorizer(authorizer Authorizer) {
```
gammazero: How do we provide different Authorizers?  I did not see that as part of the RealmConfig.

mourad: Missed that. Good catch!

aat/main_test.go
mourad: I have not used it, but that's not to say that my use case is the usual case. I didn't make a change not, but we can if we decide that that's the way to proceed.

aat/main_test.go
mourad: Done, removed the option to disable it.

router/router_config.go
mourad: moved this to `router.go`, and `RealmConfig` to `realm.go`. Alternatively, we could have created a `config.go` but I don't feel strongly at the moment either way.

PR #16, opened 2017-08-04
Channelsizing

gammazero:
This is a smaller PR in prep for some more changes to broker/dealer to move work of critical routing goroutine.

The outQueue size is no longer configurable as the size should be a small fixed size only needed in the case that kernel buffers are filled up temporarily and a send to the client blocks.  It is not helpful for the user to try to provide a guess for the best size.

### DIFF
--- a/router/router_test.go
+++ b/router/router_test.go
@@ -127,7 +127,10 @@ func TestHandshakeBadRealm(t *testing.T) {
 
 func TestRouterSubscribe(t *testing.T) {
 	const testTopic = wamp.URI("some.uri")
-	r, _ := newTestRouter()
+	r, err := newTestRouter()
+	if err != nil {
+		t.Error(err)
+	}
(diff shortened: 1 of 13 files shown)


router/router_test.go
```diff
@@ -103,32 +108,29 @@ func TestHandshakeBadRealm(t *testing.T) {
...
+		}
 	}
 }
 
 func TestRouterSubscribe(t *testing.T) {
 	const testTopic = wamp.URI("some.uri")
-
-	sub, subServer := transport.LinkedPeers(0, log)
-	r, err := newTestRouter()
+	r, _ := newTestRouter()
```
mourad: even though this generally should not fail in the test, I think it is a good idea not to ignore the err (if there is one)

gammazero: done

PR #17, opened 2017-08-05
Replace Submit() with specific functions in broker and dealer. 

gammazero:
- Reduce work done in broker and dealer critical sections.
- Do not require dealer to know about meta procedures.

Note: Dealer still needs to know about meta session so that it can publish registration meta events.


mourad:
I had a few suggestions:

- I liked the idea of a single func dealing with the Meta API requests switching on an `iota` value (similar to how it used to be) vs. a number of `func`s (one for each request type) at the dealer interface.  It also seems strange that the realm would be the one that knows how to `Yield` messages together. It seems more appropriate for the single `func` to return the appropriate `wamp.Message` and all the realm does is send it using the `metaClient`.

- To completely remove the dealer's dependency in the `metaClient`, you could modify the dealer's interface to return `*wamp.Message` or `[]*wamp.Message`.  As an alternative approach, arguably better than modifying the interface, is to allow the dealer to send the *wamp.Message over a channel to the realm to sent. This would remove the need of return values on any interface func, including the `func` that deals with the Meta API requests..

gammazero:
The `realm.metaProcMap` already maps a registration ID to a specific registration meta procedure.  To have a single dealer function handle this means re-encoding the meta procedure, and having the dealer re-decode it (using a map or switch) just to determine the function to call, which was previously done.  I think a better way to simplify the interface is to have the dealer keep track of the registered meta procedures and run the meta procedure handler.

mourad:
Looks good to me

PR #19, opened 2017-08-08
Handle dropped invocation.

### LINKED ISSUE #8: Dropped INVOCATION needs to result in ERROR to caller.
If an INVOCATION message is dropped due to client's queue being full, this should result in sending an ERROR back to the caller, since there will never be any result (WAMP spec does not dictate how to handle this other than indicating a network error).  Optionally, an unresponsive client can be dropped after some time or some number of dropped messages.

Queue size is currently configurable, but a reasonable size may not be known in advance.

Another way to handle this is to have a queue that can can keep growing, at least until resource exhaustion.  This is better for situations where reliable message delivery is most important, but risks impacting the router if the slow client cannot recover and its queue continues to grow.

gammazero:
This closes #8

### DIFF
--- a/router/dealer.go
+++ b/router/dealer.go
@@ -69,7 +69,7 @@ type Dealer interface {
 	Yield(callee *Session, msg *wamp.Yield)
 
 	// Error handles an invocation error returned by the callee.
-	Error(sess *Session, msg *wamp.Error)
+	Error(msg *wamp.Error)
 
 	// Remove a callee's registrations.
 	RemoveSession(*Session)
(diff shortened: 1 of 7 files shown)


router/dealer.go
```diff
@@ -692,43 +701,47 @@ func (d *dealer) cancel(caller *Session, msg *wamp.Cancel) {
 
 	// Cancel mode should be one of: "skip", "kill", "killnowait"
 	mode := wamp.OptionString(msg.Options, "mode")
```
mourad: seems that we should validate that it is one of the expected modes, and default to "skip" if not set.

gammazero: Done

mourad: I can't recall where I saw that 'skip' should be the default (can't find it now), but I noticed that on the client, "killnowait" is the default when none is given.

Either way, it seemed that anything other than those 3 should be an error, except when omitted should be defaulted.

gammazero: It is not really clear what the default should be, since the meaning/workflow of each mode is not specified in the spec.  Also, looks like autobahn/crossbar uses "abort" instead of "killnowait".
https://github.com/guaracheitor/AutobahnPython/blob/master/autobahn/wamp/message.py#L1790

Cannot find any documentation of options, so was intentionally not rejecting unrecognized options, and defaulting to safe behavior of skip, since that cancels at the router, but does not send an INTERRUPT message to the callee.

So, what is preferred behavior?
1) Return error to caller on CANCEL with unrecognized mode
2) Log unrecognized mode and default unrecognized mode to "skip"

mourad: Makes sense. Looks good to me.

PR #21, opened 2017-08-12
Do not make clients include transport/serialize to specify serializer

gammazero:
This gets rid of a mostly unnecessary import.  So having serialization as a separate package is not so bad.

With this change it does not seem to buy much to move the code from the serialize package into the transport package.  If you see any advantage let me know either way let me know.


mourad:
looks good to me

PR #24, opened 2017-08-13
Implementation to filter by org_id using client filter tool

gammazero:
*DO NOT MERGE*

This is to demonstrate filtering by org_id by having client examine session details to maintain whitelist of session IDs.

NOTE: Changes to session meta in `realm.go` may not comply with WAMP spec.


gammazero:
not moving forward with this approach

PR #25, opened 2017-08-15
Blackwhitelist

gammazero:
- Improved blacklist/whitelist handling so that the b/w lists are only extracted once from the publish message.
- Extend subscriber blacklist/whitelist for any string attribute match (i.e. `exclude_xxx = {val1, val2 ..}` and `eligible_xxx = {val1, val2, ..}`)
- Return all session details in session meta API on_join and get_session.

### DIFF
--- a/aat/sessionfilter_test.go
+++ b/aat/sessionfilter_test.go
@@ -29,7 +29,7 @@ func TestWhitelistAttribute(t *testing.T) {
 	}
 	sync2 := make(chan struct{})
 	evtHandler2 := func(args wamp.List, kwargs wamp.Dict, details wamp.Dict) {
-		sync1 <- struct{}{}
+		sync2 <- struct{}{}
 	}
 	err = subscriber2.Subscribe(testTopic, evtHandler2, nil)
 	if err != nil {
@@ -93,7 +93,7 @@ func TestBlacklistAttribute(t *testing.T) {
 	}
 	sync2 := make(chan struct{})
 	evtHandler2 := func(args wamp.List, kwargs wamp.Dict, details wamp.Dict) {
-		sync1 <- struct{}{}
+		sync2 <- struct{}{}
 	}
 	err = subscriber2.Subscribe(testTopic, evtHandler2, nil)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


aat/sessionmeta_test.go
```diff
@@ -48,6 +48,16 @@ func TestMetaEventOnJoin(t *testing.T) {
...
 
+	// Wait for any event from subscriber joining.
+	var timeout bool
+	for !timeout {
+		select {
+		case <-errChan:
+		case <-time.After(200 * time.Millisecond):
+			timeout = true
+		}
+	}
```
mourad: Could you explain the reason for this ?

aat/sessionfilter_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	// Setup subscriber2
+	subscriber2, err := connectClientDetails(wamp.Dict{"org_id": "other"})
+	if err != nil {
+		t.Fatal("Failed to connect client:", err)
+	}
+	sync2 := make(chan struct{})
+	evtHandler2 := func(args wamp.List, kwargs wamp.Dict, details wamp.Dict) {
+		sync1 <- struct{}{}
```
mourad: this too.

aat/sessionfilter_test.go
```diff
@@ -0,0 +1,136 @@
...
```
mourad: this should be sync2.

aat/sessionmeta_test.go
ajgillis: This is used to catch and discard the event published when the subscriber joins the realm, if that event has not yet been published before the subscription is created.  The test discards this, if it happens, because it is not reliable.  The test is interested in checking for the event triggered by another session's subscription.

aat/sessionfilter_test.go
ajgillis: Done

aat/sessionfilter_test.go
ajgillis: Done

PR #26, opened 2017-08-16
Missed publishing onUnsubscribe event. 

gammazero:
- Do not send meta event to session causing meta event to be generated.
- Add AATs for subscription meta events.


mourad:
LGTM

PR #28, opened 2017-08-16
Call event handlers in order received

gammazero:
When testing pub/sub with the nexus client, I often saw the subscribers' event handlers getting called out of order.  I verified the client was always receiving the events in the correct order.  The problem occurred because the client ran the event handlers concurrently and could not guarantee they would be run in the same order.

While this is not necessarily a defect, and may offer better performance for some clients, it would likely be unexpected by most users of the nexus client, which could lead to problems.

Please consider whether this fix is good/sufficient, or if you have use cases that demand an option to run event handlers concurrently.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -732,16 +732,27 @@ func (c *Client) receiveFromRouter() {
 	}
 }
 
+// handleEvent calls the event handler function a subscriber designated for
+// handling EVENT messages.
+//
+// The eventHandlers are called serially so that they execute in the same order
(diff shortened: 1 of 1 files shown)


mourad:
LGTM.

I also had thought that running concurrently was a better approach, however, as you point out, it could be unexpected for most clients.  If the user really does want/need the handler to run concurrently, they can simply start a goroutine from within the EventHandler and return control right away to the nexus client.

PR #29, opened 2017-08-16
Option to match subscribe options with white/blacklisting.

gammazero:


gammazero:
Not moving forward with this approach.

PR #31, opened 2017-08-21
Shutdown

gammazero:
I believe this fixes the possible send on closed channel error, which was do to prematurely closing the realm's action channel.

The realm shutdown procedure has also been clarified.

Meta events are not generated during shutdown.  There is no point to generating meta events at realm shutdown since those events would only be received by meta event subscribers that had not been removed yet, and clients are removed in any order.

### DIFF
--- a/aat/sessionmeta_test.go
+++ b/aat/sessionmeta_test.go
@@ -78,10 +78,6 @@ func TestMetaEventOnJoin(t *testing.T) {
 		t.Fatal(metaOnJoin, "meta even had wrong session ID")
 	}
 
-	err = subscriber.Unsubscribe(metaOnJoin)
-	if err != nil {
-		t.Fatal("unsubscribe error:", err)
-	}
 	err = subscriber.Close()
 	if err != nil {
 		t.Fatal("Failed to disconnect client:", err)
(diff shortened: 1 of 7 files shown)


mourad:
It doesn't seem right to leak realm state (`shutdown` information) into the broker/dealer to prevent them from sending meta information. It seems that this could be prevented by waiting until the broker/dealer are completely shutdown before closing the meta session.

I can't reason at the moment about sending on a closed channel you found, but the one that I hit was here:
https://github.com/gammazero/nexus/blob/master/client/client.go#L780-L783

gammazero:
The problem is I cannot close the broker and dealer until all the sessions are closed, in order to ensure that no additional messages will be sent.  This includes the metadata session, which may have a message waiting to be sent if its goroutine has not been scheduled yet.  If broker closes before meta data session, then it risks send on closed channel (or deadlock if channel left open, but not serviced).

The solution is to simply not remove sessions from broker and dealer during shutdown, and that certainly prevents meta events when closing client sessions without leaking realm state.

I will track down the client issue, thank you.

gammazero:
For the client issue, it looks like there needs to a wait for any active event handlers, similar to what is done for invocation handlers:
https://github.com/gammazero/nexus/blob/master/client/client.go#L637-L643

aat/sessionmeta_test.go
```diff
@@ -264,13 +262,17 @@ func TestMetaProcSessionCount(t *testing.T) {
 		t.Fatal("Session count should be same as first")
 	}
 
-	err = caller.Close()
+	subscriber.Unsubscribe(metaOnJoin)
+	subscriber.Unsubscribe(metaOnLeave)
```
mourad: Is it a requirement to unsubscribe before closing?

gammazero: Not a requirement at all.  I guess there is no need, especially since I am not checking for error on unsubscribe.

PR #36, opened 2017-08-26
Joining a realm is part of client creation.

gammazero:
This simplifies the code both inside client and outside in the use of the client.

- Client configuration makes more sense.
- Simpler for users to create clients and inspect errors.
- Removes internal and external complications of having a client that may or may not have joined a realm.
- Client shutdown logic much simpler.


mourad:
LGTM

PR #37, opened 2017-08-28
Minor enhancements

mourad:


realm.go
```diff
@@ -140,14 +140,12 @@ func (r *realm) close() {
 	// messages can be generated once sessions are closed.
 	r.waitHandlers.Wait()
 
-	if r.metaSess != nil {
```
gammazero: The reason to check `if r.metaSess != nil` is in case the realm was closed (during router shutdown), and a realm had been added, but the goroutine to run it had not yet been scheduled.  In that that edge case, `realm.close()` would be called, but `r.metaSession` would be nil.

gammazero: ... or course it is also as bad if the goroutine does get scheduled just after the check is done because then the meta handlers get leaked.

mourad: sounds like there's still concern about concurrency issues during shutdown. we'll need to check/simplify it, but I don't that's the reason to check that here (it should be prevented).

gammazero: I agree, if that check is needed, then that is a problem.

PR #38, opened 2017-08-28
Protect against goroutine leak if router stops reading messages...

gammazero:
... from client before client is done sending.

During router shutdown, if the router stops listening to a client, but the client tries to send a message before it receives the shutdown signal, its receiveHandler may hang trying to write the message to a channel that the router is no longer reading.  If this happens, the channel must be closed to free the goroutine, and the resulting panic must be recovered.

The websocket peer can, but probably should not, just close the channel, because there may be some messages to route, or at least a GOODBYE to send before closing.

### DIFF
--- a/transport/websocketpeer.go
+++ b/transport/websocketpeer.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"runtime"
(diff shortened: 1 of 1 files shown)


gammazero:
no, no, no

PR #41, opened 2017-09-08
Add RegistrationID client API.  Add registration meta API tests.

gammazero:


client/client.go
```diff
@@ -246,14 +246,22 @@ func (c *Client) Subscribe(topic string, fn EventHandler, options wamp.Dict) err
...
 func (c *Client) SubscriptionID(topic string) (subID wamp.ID, ok bool) {
 	subID, ok = c.topicSubID[topic]
 	return
 }
 
+// RegistrationID returns the registration ID for the specified procedure.  If
+// the client is not registered for the procedure, then returns false for
+// second boolean return value.
+func (c *Client) RegistrationID(procedure string) (regID wamp.ID, ok bool) {
+	regID, ok = c.nameProcID[procedure]
```
mourad: this is a problem for concurrent calls, same in case of `SubscriptionID()`

gammazero: fixed

PR #46, opened 2017-09-14
Authorize() returns optional reason string for client

gammazero:
If `Authorize()` returns false, an optional reason string may be returned to provide additional information to the client about the unauthorized response, or the authorization failure.

This is a API change, so need to consider carefully.

### DIFF
--- a/aat/auth_test.go
+++ b/aat/auth_test.go
@@ -136,7 +136,7 @@ func TestAuthz(t *testing.T) {
 	if err == nil {
 		t.Fatal("Expected error")
 	}
-	if !strings.HasSuffix(err.Error(), "wamp.error.not_authorized: Cannot contact LDAP server") {
+	if !strings.HasSuffix(err.Error(), "wamp.error.authorization_failed: Cannot contact LDAP server") {
 		t.Fatal("Did not get expected error message with reason, got:", err)
 	}
 }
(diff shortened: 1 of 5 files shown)


mourad:
Instead of changing the interface to return an optional string, have you considered returning a specific `error` that could carry the reason itself?

gammazero:
Currently, the presence of error is used to differentiate between not authz and failed authz.  If I understand what you are suggesting it is to have `Authorize()` return an `AuthzError` type of error (and type assert the error into that type).  The AuthzError would contain both the error type (`not_authorized`, or `authorization_failed`) and an optional message?

In the case of authz failure, the error may contain information that gets logged, but is not meant for the client (i.e. info concerning the inner workings of the authz provided).  The optional reason message is one that is meant for the client specifically, so the `AuthzError` would  need to contain 3 items:
1. error type
2. error message
3. optional reason message

This would make the boolean returned by `Authorize()` unnecessary, since that could now be determined by `if err != nil`

gammazero:
It may be a bad thing to encourage sending reason info back in a `not_authorized` response, since that risks disclosing information that could help an attacker's efforts to probe (e.g. "this user is not an admin").  Maybe better to limit returning error information to `authorization_failed` response, since that is a case where authorization could not be determined (e.g. could not connect to policy DB).

mourad:
I agree that `authorization_failed` would be a better place to return a reason.

mourad:
LGTM

PR #47, opened 2017-09-17
RawSocket transport

gammazero:
Implement RawSocket transport.

- Add option to run AATs with Unix or TCP rawsocket.
- When creating a rawsocket client or server, optionally specify maximum message size to receive.


gammazero:
Need to rework WebsocketServer interface

PR #51, opened 2017-10-03
New client creation API

gammazero:
This was an experiment to simplify the client creation api.   All of the `NewXxxClient()` functions have been replaced by a `ConnectXxx()` function.  The main difference is that almost all of the parameters have been put into the `ClientConfig` structure.  This makes it simpler to provide default values and makes all of the client creation APIs more similar.

Let me know if this seems like an improvement, or if you would prefer some changes.


gammazero:
A `client.Connect(cfg)` would not be able to differentiate between a websocket or tcp raw socket connection based on config or address.  So, unless I add a parameter I still need `ConnectXxx` where `Xxx` is `Websocket`, `TCP`, or `Unix`.

The reason I did not put the address into cfg is because the address is not used by the client, but only by the transport to create a connected `Peer`.  If a client provided a different transport/peer implementation, then the address would have no use in the cfg.

I will update `client.NewClient` not to take a `logger` parameter.  That does constitute an API change, but I doubt anyone is using that, so maybe I can still get away with a minor version change.  Or, is it better to leave it and then remove it when I do a major version change, where I also delete the deprecated functions?

mourad:
I think the `client.Connect(cfg)` could determine the correct transport by parsing the address with `url.Parse`, and looking at the `Scheme`. Something similar to:
https://github.com/fsouza/go-dockerclient/blob/master/client.go#L961-L1000

My opinion is that it would be ok to include the `address` in the `cfg` in this PR even though it would not be used when calling `client.NewClient` with a connected peer already. The same, if I'm not mistaken, is true with regards to `Serialization`, `TlsConfig`, `Dial` and `RecvLimit`. I don't believe there's an overlap between the transport parameters and the client parameters (except for `logger`) so, alternatively, you could create a `TransportConfig` and call `client.Connect(transportConfig, clientConfig)`.

Another option could be to create a `Transport` nested struct in `ClientConfig` that would be completely ignored when calling `client.NewClient`.

With regards to the `logger` parameter, not sure what is better.  An argument is to not including the `logger` into the `ClientConfig`, so there's no overlap between ClientConfig and the transport parameters. This way the API doesn't have to change either. In which case, the previous comment, would be `client.Connect(transportConfig, clientConfig, logger)`.

gammazero:
To determine protocol from address requires a URL, at least for websocket, `ws://host/`.  If I do that, then should every address be in URL form: `tcp://host/`, `unix:///path/to/socket`?  I could opt to include a `network` parameter in the cfg and it could have a value of `"ws"`, `"tcp"`,  or `"unix"` and define consts for those values.  I avoid all that by making the transport explicit in the name of the function.  The function name may also help to avoid using the wrong transport without noticing, with a WAMP server that supports multiple transports. 

You are absolutely correct about the address parameter, it is part of the transport config as is serialization, tlscfg, etc.  So, really there are two sets of config params: transport config and client config.  Well, three, considering logger as separate.  So, I am not sure which is preferable (probably subjective).  Users, who are not terribly invested in understanding finer points of parameter classification, want the clearest/simplest way to create a client.  Here are examples of different organizations of the parameters:

---- ALL-IN-ONE ----
```
cfg := client.ClientConfig{
    Address:       "myhost.mydomain.net",
    Realm:         "nexus.examples",
    TlsCfg:        tlscfg,
    Serialization: serialization,
    Logger:        logger,
}
cli, err := client.ConnectWebsocket(cfg)
```

---- TWO-CONFIGS ----
```
tCfg := client.TransportConfig{
    Address:     "myhost.mydomain.net",
    TlsCfg:        tlscfg,
    Serialization: serialization,
}
cCfg := client.ClientConfig{
    Realm:   "nexus.examples",
}
cli, err := client.ConnectWebsocket(tCfg, cCfg, logger)
```

Could also have a separate config type for each transport.
```
tCfg := client.WebsocketConfig{
    Address:     "myhost.mydomain.net",
    TlsCfg:        tlscfg,
    Serialization: serialization,
}
cCfg := client.ClientConfig{
    Realm:   "nexus.examples",
}
cli, err := tCfg.Connect(cCfg, logger)
```

---- NESTED ----
```
cfg := client.ClientConfig{
    Realm:   "nexus.examples",
    TransportConfig: {
        Address:     "myhost.mydomain.net",
        TlsCfg:        tlscfg,
        Serialization: serialization,
    }
}
cli, err := client.ConnectWebsocket(cfg, logger)
```
 
I think the first, all-in-one, would be preferable to most users.  If a client has their own transport/peer, then they will just have to know that NewClient ignores the transport params.

I could pull `Address` and `Realm` out of the config and make them explicit parameters, and let cfg contain all the optional parameters, and allow a `nil` value to be passed.
```
cli, err := client.ConnectWebsocket("myhost.mydomain.net", "nexus.examples", logger, nil)
```

or, drop PR if it does not actually help overall.

mourad:
I don't think you should drop of the PR.  What I was suggesting is that I think you could add a top-level `client.Connect()` function that can determine which specific call to make based on protocol. I'm thinking users might have to perform the same sort of logic that is in `main_test.go`, so instead, it could be encapsulated into the code and users can just rely on that, and we can extend it should nexus decide to support new protocols.

So you would end up with:
`client.NewClient(p wamp.Peer, cfg ClientConfig, logger stdlog.StdLog)`
and
`client.Connect(tcfg TransportConfig, ccfg ClientConfig, logger stdlog.StdLog)` or perhaps `client.NewClientConnect(...)` would make it clearer that it returns a client after connecting?

After taking a second look, the value that `ConnectXxx` or even the previous `NewXxxClient` provided was a way to start up a specific transport connection and return a connected client.  If we provide a specific `TransportConfig` and a way that the client can determine how to connect, I wonder if the various APIs to get a client are still needed? Of course, the user could still decide to get a peer directly through `transport` and pass it into `client.NewClient` if they choose.

You are correct that the best approach is definitely subjective. I am leaning towards separate configs, which would then lead to possibly providing `transport.NewPeer(cfg TransportConfig)` to simplify the creation of a peer.  If we do this, then the creation of the transport is in the `transport` package using the TransportConfig, and all that `client.Connect` or `client.NewClientConnect` would do is wrap the call to it and return a client.

I am not against providing the URL in the config such as `ws://`, `tcp://` and `unix://`. You could provide it as a separate parameter in the config if you prefer. One thing that I missed in my previous review is that `client.NewWebsocketClient` used to take the URL, but now it just takes the address. With this change it's impossible to connect with `wss://` with the helper functions and would require the user to use the transport directly.

gammazero:
After trying every combination of parameter organization that we discussed, including address parsing to determine transport, I have arrived at a conclusion of the way I think works best for the API:

1. Transport is explicit in function name: ConnectWebsocket(), ConnectTCP(), ..
2. Address is not part of config: ConnectWebsocket("10.20.30.42:8080", cfg)
    - No URL schemes, no address parsing
    - Address tells where to go, cfg tells how to communicate
    - Easy to see if address looks correct for transport
    - Local and Socket APIs more consistent (first param is what to connect to)
3. Decided to put logger in cfg, so it can be ignored when default value is ok.

mourad:
Seems that support for `wss` is still missing in `ConnectWebsocket`.

Does `ConnectTCP` need to support `tcp4` and `tcp6` explicitly or is that automatically supported by the underlying lying library when using `tcp`?

gammazero:
The two-config implementation ends up as not much different than what it was before this PR.

The `NewXxxClient()` was just a wrapper for `transport.ConnectXxxPeer()` + `client.NewClient()` and was there mostly to prevent clients from having to import `transport`.  Having a transport config in the transport package now makes clients import  `transport`, so not much need for `NewXxxClient()` or `ConnectXxx()` APIs.  Therefore, two-config approach converges to:
`transport.ConnectXxxPeer()` + `client.NewClient()` with a cfg instead of a few params when creating the peer.

gammazero:
ConnectWebsocket does not use URL scheme.  The address must contain the port (like ConnectTCP).  ws vs wss is determined by presence of tcl.Config.  Sounds like I either need to document more clearly, or if people expect to give URL, then I need to accept that.  Autobahn client seems to use URL: `wss://addr/ws`.  Not sure what the significance of the trailing `/ws` is.

ConnectTCP automatically supports both tcp4 and tcp6 depending on the address given (tested).

mourad:
Looking at the vendored github.com/gorilla/websocket implementation, decision whether to use TLS is determined by the scheme, not by the presence of `tls.Config`:

https://github.com/gammazero/nexus/blob/master/vendor/github.com/gorilla/websocket/client.go#L185-L192

Omitting `tls.Config` while using `wss` results in the default `tls.Config`:
https://github.com/gammazero/nexus/blob/master/vendor/github.com/gorilla/websocket/client.go#L326
https://github.com/gammazero/nexus/blob/master/vendor/github.com/gorilla/websocket/client_clone.go#L12-L14

gammazero:
The one way that does simplify things is to force a URL scheme to be used for all addresses.  Then there is no ambiguity in address parsing, and there can be a single `client.Connect()`
URLs could be like this:
Websocket:  `ws://addr/` or `wss://addr/`
TCP: `tcp://addr/` or `tcps://addr/`
Unix: `unix://path`

`tcps` seems weird though

gammazero:
I went with "tcp://" and "tcps://" for consistency with "ws://" and "wss://" when determining whether or not to use TLS.

mourad:
looks good 👍

Issue #57, opened 2017-10-10
Move router code into router package.


gammazero:
Currently all the router code is in the project root directory.  The client code is in its own directory/package as are some packages common to both.  Seems like the code would be better organized if all of the router-specific code was under a "router" directory.

If the router code were moved into a different package, this would break backwards compatibility and require a major version change (nexus-2.0).

Does it make sense to move the router code?

gammazero:
Resolved by PR #59

PR #59, opened 2017-10-17
Move all router realated code into router package.  Addresses #57.

gammazero:
Do not merge yet.  Please consider this reorganization of the source.


mourad:
I do find this layout easier, but that's just my opinion.

I know you were concerned about what this change would mean to existing code, so just wondering, did you consider type aliases in go 1.9?

gammazero:
No, but will see what that looks like when I have time.  Thanks for the suggestion.

gammazero:
Aliasing works great, but then I make the code incompatible for anything less then Go 1.9.  Maybe better just to make a clean break and remove the deprecated APIs.

PR #64, opened 2017-10-29
Explicit blocking or non-blocking send

gammazero:


transport/localpeer_test.go
```diff
@@ -40,31 +37,30 @@ func _TestSendRecv(t *testing.T) {
...
+		err = r.TrySend(&wamp.Publish{})
 		close(done)
 	}()
 	select {
 	case <-done:
 	case <-time.After(time.Second):
 		t.Fatal("Send should have dropped and not blocked")
 	}
-	if err == nil || !strings.HasPrefix(err.Error(), "client blocked") {
+	if err == nil || err.Error() != "try again" {
```
mourad: prefer a typed error instead of a string comparison.

router/broker.go
```diff
@@ -555,3 +554,12 @@ func (b *Broker) pubSubCreateMeta(subTopic wamp.URI, subSessID, subID wamp.ID, m
 	}
 	b.pubMeta(wamp.MetaEventSubOnCreate, sendMeta)
 }
+
+func (b *Broker) trySend(sess *wamp.Session, msg wamp.Message) error {
```
mourad: instead of having this function in both, the `broker` and the `dealer` just for logging purpose, have you considered allowing the various `broker` and `dealer` methods return the error and handle the logging in the `realm`?

The error from `trySend` is just being thrown away anyway.

transport/localpeer_test.go
gammazero: I would agree if there was a general need to branch on error in using this API.  However, this is unit test code that just makes sure the correct error is returned.  I considered returning `syscall.EAGAIN` aka`syscall.EWOULDBOCK`.

router/broker.go
gammazero: Problem with returning the message is that it would need to be returned via a channel to the realm (or session in realm).  Broker or dealer cannot call realm either since they do not know about the parent realm.   Still, yes, there probably is a better way that having the same code in two places.

PR #75, opened 2018-01-31
DynamicRouter Interface

dcarbone:
Hey there.

First, thanks for creating this package!  I'm grateful somebody else took up the Go WAMP reigns with the lack of work on Turnpike :)

One feature I have a need for, though, is the ability to dynamically add and remove Realms with a specific configuration.  The "Template" mechanism does not work for my use-case as I will need to create Realm-specific Authenticators and Authorizer instances.  

To that end, I added the interface `DynamicRouter` and modified the default router slightly to support this new interface.  

Let me know what you think!

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -325,20 +325,31 @@ func (r *router) AddRealm(config *RealmConfig) error {
 
 // RemoveRealm will close and then remove a realm from this router, if the realm exists.
 func (r *router) RemoveRealm(name wamp.URI) {
+	// Because we want to force atomicity as briefly as possible, the atomic func will be used purely to attempt to
+	// locate the realm
+	var realm *realm
+	var ok bool
 	sync := make(chan struct{})
 	r.actionChan <- func() {
-		if realm, ok := r.realms[name]; ok {
-			realm.close()
+		if realm, ok = r.realms[name]; ok {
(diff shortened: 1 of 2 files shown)


gammazero:
I like this, and the implementation looks good.  Before I merge it, I need to work through a few edge cases such as what happens when a client is joining a realm, and the realm gets removed.

router/router.go
```diff
@@ -51,6 +51,17 @@ type Router interface {
...
+// An DynamicRouter is a router that allows for dynamic Realm adding.
+type DynamicRouter interface {
+	Router
+
+	// AddRealm will append a realm to this router
+	AddRealm(*RealmConfig) error
+
+	// RemoveRealm will attempt to remove a realm from this router
+	RemoveRealm(wamp.URI)
+}
```
gammazero: Thinking we should just add `AddRealm` and `RemoveRealm` to `Router`.  I do not think a new interface is needed.

gammazero: Please combine.  Weighing the risk of breaking bc versus what looks right here, I choose to go with what results in "better" code.  Submit change, and I I'll do a final review and commit.  Thank you.

dcarbone:
Interfaces merged

PR #80, opened 2018-02-19
Use hmac.Equal to check signature

azamat-sharapov:


router/auth/crauth.go
```diff
@@ -87,7 +88,8 @@ func (cr *CRAuthenticator) Authenticate(sid wamp.ID, details wamp.Dict, client w
 	}
 
 	// Check signature.
-	if crsign.SignChallenge(chStr, key) != authRsp.Signature {
+	signedCh := crsign.SignChallenge(chStr, key)
+	if !hmac.Equal([]byte(authRsp.Signature), []byte(signedCh)) {
```
gammazero: `hmac.Equal` is meant for comparing _unencoded_ hmac values.  The code above is comparing base64-encoded hmac values, which is not the correct usage of `hmac.Equal`.

To correctly use `hmac.Equal`, `crsign.SignChallenge` should not base64-encode the signature and `authRsp.ignature should also be raw bytes, or these need to be base64-decoded first.

The reason the returned signature is base64 encoded is because that is how it is shown in the WAMP spec: http://wamp-proto.org/static/rfc/draft-oberstet-hybi-crossbar-wamp.html#rfc.section.14.5.2.3

After computing the signature, the client will send an AUTHENTICATE message containing the signature:
```
    [5, "gir1mSx+deCDUV7wRM5SGIn/+R/ClqLZuH4m7FJeBVI=", {}]
```

gammazero:
I made the changes here in PR #81 .  Created a `crsign.VerifySignature` so that importing `"crypto/hmac"` is not needed elsewhere.  Comment gives credit to you for this.

Thank you for this contribution.

azamat-sharapov:
@gammazero sorry for late response, I could have been more useful if read your comment earlier. Thanks for solution and the credit :)

PR #88, opened 2018-03-20
Cookieauth

gammazero:
In the `WebsocketConfig` used to configure the router, there are two new boolean values: `EnableTrackingCookie` and `EnableRequestCapture`

`EnableTrackingCookie` tells the server to send a random-value cookie to the websocket client.  A returning client may identify itself by sending a previously issued tracking cookie in a websocket request.  If a request header received by the server contains the tracking cookie, then the cookie is included in the HELLO and session details.  The new tracking cookie that gets sent to the client (the cookie to expect for subsequent connections) is also stored in HELLO and session details.                                

The cookie from the request, and the next cookie to expect, are stored in the HELLO and session details, respectively, as:
 ```                                        
 Details.transport.auth.cookie|*http.Cookie                           
 Details.transport.auth.nextcookie|*http.Cookie
```                       
                                                                          
This information is available to auth/authz logic, and can be retrieved from details as follows:                                                 
```
req *http.Request
v, err := wamp.DictValue(details, []string{"transport", "auth", "request"})                              
if err == nil {
    req = v.(*http.Request)
}
```
The "cookie" and "nextcookie" values are retrieved similarly. 

`EnableRequestCapture` tells the server to include the upgrade HTTP request in the HELLO and session details.  It is stored in `Details.transport.auth.request|*http.Request` and is available to auth/authz logic.

### DIFF
--- a/aat/auth_test.go
+++ b/aat/auth_test.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"crypto/sha256"
 	"errors"
+	"fmt"
+	"net/http"
+	"net/http/cookiejar"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

**Comment by gammazero** [comment]
Note: `session.Details.trasnport.auth` information is not available in the session meta info because:
1. It is not serializable (stored as go http objects)
2. Prevent exposure of authentication material

File: aat/auth_test.go [hunk 1]
```diff
@@ -192,3 +248,45 @@ func (ks *serverKeyStore) AuthRole(authid string) (string, error) {
...
+
+	v, err = wamp.DictValue(details, []string{"transport", "auth", "cookie"})
+	if err != nil {
+		// No tracking cookie - client not recognized, so not auth.
+		return false
+	}
+	cookie := v.(*http.Cookie)
+
+	// Tracking cookie matches cookie of previously good client.
+	if cookie.Value != ks.cookie.Value {
```
azamat-sharapov: @gammazero could you please, clarify this part? When there are multiple connections, is KeyStore instance created per each client?

gammazero: A KeyStore is provided to an authenticator instance, which is registered with the router.  All connections using the same authenticator will use the same KeyStore.  If you are trying to determine which cookie to retrieve for comparison, you need to look at authid or other information in the HELLO.Details (possibly even request data).

Each connection is given a tracking cookie with a unique value.  If the same client makes a connection in the future it can identify itself with a previously issued tracking cookie.  This means that the authenticator needs to store the nextCookie when the client is authenticated.  Then, in the future, the authenticator can see if the saved cookie matches the one presented by the client.  If the client is authenticated, then store the new nextCookie.

azamat-sharapov: > The test just looks at a single saved cookie to make sure stuff works.

Ah, sorry, completely forgot this was a test file :D 

So this line is meant for comparing stored cookie with the one from request. Thanks!

PR #95, opened 2018-05-03
Implement CBOR serialization, replace JSON serializer

martin31821:
This commit replaces the golang `encoding/json` by `ugorji/co/codec` for speed.
Additionally, I implemented a CBOR serializer and deserializer and added tests
to ensure it's working as expected.

Binary compatibility to other routers such as crossbar.io should be given, since
the implementation is using the default cbor name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,8 @@ test:
 	go test -race ./aat -scheme=unix
 	go test ./aat -scheme=ws -serialize=msgpack
 	go test ./aat -scheme=tcp -serialize=msgpack
+	go test ./aat -scheme=ws -serialize=cbor
+	go test ./aat -scheme=tcp -serialize=cbor
 	go test ./aat -scheme=wss
 	go test ./aat -scheme=tcps
 
--- a/aat/main_test.go
+++ b/aat/main_test.go
@@ -77,11 +77,11 @@ func TestMain(m *testing.M) {
 	flag.StringVar(&scheme, "scheme", "",
 		"-scheme=[ws, wss, tcp, tcps, unix] or none for local (in-process)")
 	flag.StringVar(&serType, "serialize", "",
-		"-serialize[json, msgpack] default is json")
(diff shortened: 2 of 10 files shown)


gammazero:
Wow, this is awesome!  I will review and test as soon as I have a chance this weekend (maybe sooner).

It looks like your changes to `transport/websocketpeer.go` are on a previous version of that file.  Let me know if you can take a look at that and merge it with the latest.

Thank you so much for this.  It was something I have been wanting in nexus for some time, and this looks great.

martin31821:
Yeah of course. I originally started hacking on this within an application where v2.0.2 was vendored and after I finished I just copied that over.

Will take a look.

martin31821:
@gammazero Updated the file

gammazero:
Testing all looks good.  Just a few small changes to finish replacing std json codec:
Here: https://github.com/gammazero/nexus/pull/95/files#diff-f35b5582ea5620327d0b6ac0c11883c9R60
replace
```
    return json.Marshal("\x00" + s)
```
with
```
    var out []byte
    jsh := &codec.JsonHandle{}
    return out, codec.NewEncoderBytes(&out, jsh).Encode("\x00" + s)
```

Here: https://github.com/gammazero/nexus/pull/95/files#diff-f35b5582ea5620327d0b6ac0c11883c9R65
replace
```
    err := json.Unmarshal(v, &s)
```
with
```
    jsh := &codec.JsonHandle{}
    err := codec.NewDecoderBytes(v, jsh).Decode(&s)
```

and then you can remove this: https://github.com/gammazero/nexus/pull/95/files#diff-f35b5582ea5620327d0b6ac0c11883c9L5

If you want to make those changes I will merge this PR.  Also, I will vendor in a newer version of github.com/ugorji/go/codec since that has even better performance and a few fixes.

gammazero:
Added those last couple items. Will update codec in separate PR.  Thank you for contributing these changes.

PR #101, opened 2018-05-21
Add more fields when disclosing publisher or caller

martin31821:
This further improves the compatibility to crossbar.io by disclosing
caller_authid, caller_authrole, caller_authmethod and caller_authextra if set.

For pubsub, the same properties are exposed.

### DIFF
--- a/router/broker.go
+++ b/router/broker.go
@@ -457,7 +457,7 @@ func (b *Broker) pubEvent(pub *wamp.Session, msg *wamp.Publish, pubID wamp.ID, s
 		}
 
 		if disclose && sub.HasFeature(roleSub, featurePubIdent) {
-			details[rolePub] = pub.ID
+			disclosePublisher(pub, details)
 		}
 
 		// TODO: Handle publication trust levels
@@ -561,4 +561,20 @@ func (b *Broker) trySend(sess *wamp.Session, msg wamp.Message) bool {
 		return false
 	}
 	return true
+}
+
+func disclosePublisher(pub *wamp.Session, details wamp.Dict) {
+	details[rolePub] = pub.ID
+	features := []string{
+		"authrole",
+		"authid",
+		"authmethod",
+		"authextra",
+	}
+	for _, f := range features {
+		val, ok := pub.Details[f]
+		if ok {
+			details[fmt.Sprintf("%s_%s", rolePub, f)] = val
+		}
+	}
 }
--- a/router/broker.go
+++ b/router/broker.go
@@ -568,8 +568,6 @@ func disclosePublisher(pub *wamp.Session, details wamp.Dict) {
 	features := []string{
 		"authrole",
 		"authid",
-		"authmethod",
-		"authextra",
 	}
 	for _, f := range features {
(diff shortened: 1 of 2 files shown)


gammazero:
Thank you for this.  However, I must request that only `authid` and `authrole` are supported.

The WAMP standardization group has reached the consensus that the supported fields will be `authid` and `authrole`.   The discussion of this feature is here: https://github.com/wamp-proto/wamp-proto/issues/57

Some background (from discussion): On the subject fo including these fields with publisher and caller identification, according to [Tobias Oberstein](https://github.com/oberstet) (Crossbar.IO):

> Crossbar.io had implemented that in the past. And we removed it again. The reason: the next thing after authid/authrole asked for was transport level details: "we need access to the WebSocket cookie in callees". CB has more information about the Caller than just session ID, authid and authrole. Where do we stop? It doesn't make sense to transmit all that information in each and every call.

Also from Oberstein:

> This is how it is implemented in Crossbar.io now:
>
>Both publisher and caller disclosure can be enabled on a per URI basis in the router configuration
>The detail attributes in EVENT messages are: publisher (the WAMP session ID of the publisher), >publisher_authid and publisher_authrole
>The detail attributes in the INVOCATION messages are: caller (the WAMP session ID of the caller), caller_authid and caller_authrole
>I do think this reflects the end result of our discussion of the feature. @mbonneau does it?
>
>It isn't reflected in the specification though. And eg Thruway uses different attribute names

Related issue with autobahn-js: https://github.com/crossbario/autobahn-js/issues/219
and with Thruway: https://github.com/voryx/Thruway/issues/206

If we want to include additional fields, these should be optional to enable, default disabled, and the non-standard field names should begin with `x_nx_`  I would not want to do that as part of this PR, if at all.  Implementations that include additional fields are already updated or going to be updated to only include those fields as `x_...` custom fields.

Conclusion: Nexus needs to include `authid` and `authrole`, for publisher and caller identification, even though this is not yet in the WAMP spec.  It should not include additional fields as those will not be in compliance with WAMP-protocol.

router/broker.go
```diff
@@ -562,3 +562,19 @@ func (b *Broker) trySend(sess *wamp.Session, msg wamp.Message) bool {
...
 	return true
 }
+
+func disclosePublisher(pub *wamp.Session, details wamp.Dict) {
+	details[rolePub] = pub.ID
+	features := []string{
+		"authrole",
+		"authid",
+		"authmethod",
+		"authextra",
```
gammazero: Only features needed are "authid" and "authrole".  See PR discussion.

router/dealer.go
```diff
@@ -1093,3 +1093,19 @@ func (d *Dealer) trySend(sess *wamp.Session, msg wamp.Message) bool {
...
 	return true
 }
+
+func discloseCaller(caller *wamp.Session, details wamp.Dict) {
+	details[roleCaller] = caller.ID
+	features := []string{
+		"authrole",
+		"authid",
+		"authmethod",
+		"authextra",
```
gammazero: Only features needed are "authid" and "authrole".  See PR discussion.

PR #103, opened 2018-06-04
Implement testament meta api

martin31821:
This PR implements the testament meta API. 
I'm currently looking into writing some tests for this, will update the PR and remove WIP when it's done.

### DIFF
--- a/aat/sessiontestament_meta_test.go
+++ b/aat/sessiontestament_meta_test.go
@@ -1,0 +1,146 @@
+package aat
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/fortytw2/leaktest"
+	"github.com/gammazero/nexus/wamp"
+)
+
+const (
+	metaAddTestament  = string(wamp.MetaProcSessionAddTestament)
+	metaFlushTestaments = string(wamp.MetaProcSessionFlushTestaments)
+
+)
+
+func TestAddTestament(t *testing.T) {
+	defer leaktest.Check(t)()
+	// Connect subscriber session.
+	subscriber, err := connectClient()
+	if err != nil {
+		t.Fatal("Failed to connect client:", err)
+	}
+
+  errChan := make(chan error)
+  evtHandler := func(args wamp.List, kwargs wamp.Dict, details wamp.Dict) {
+    if len(args) == 0 {
+      errChan <- errors.New("missing argument")
+      return
+    }
+    _, ok := wamp.AsString(args[0])
+    if !ok {
+      errChan <- errors.New("Argument was not string")
+      return
+    }
+    errChan <- nil
+  }
+
+  err = subscriber.Subscribe("testament.test", evtHandler, nil)
+  if err != nil {
+    t.Fatal("subscribe error:", err)
+  }
+  sess, err := connectClient()
+  if err != nil {
+    t.Fatal("Failed to connect client:", err)
+  }
+  _, err = sess.Call(context.Background(), metaAddTestament, wamp.Dict{}, wamp.List{
+    "testament.test",
+    wamp.List{
+      "foo",
+    },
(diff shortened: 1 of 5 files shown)


router/realm.go
```diff
@@ -291,6 +309,27 @@ func (r *realm) onLeave(sess *wamp.Session, shutdown bool) {
...
+				r.metaPeer.Send(&wamp.Publish{
+					Request:     wamp.GlobalID(),
+					Topic:       testament.topic,
+					Arguments:   testament.args,
+					ArgumentsKw: testament.kwargs,
+					Options:     testament.options,
+				})
+			}
+			delete(r.testaments, sess.ID)
+		}
```
gammazero: Since `testaments.detached` and `testaments.destroyed` are not slices of pointers, the loops above will copy the testament struct each iteration.  Instead, use an index to access each slice element.  Indexing within a loop is highly optimized in go. Also, since the loops are identical, except for the slice they iterate, use a function so that the code does not repeat:
```go
sendTestaments = func(testaments []testament) {
    for i := range testaments {
        r.metaPeer.Send(&wamp.Publish{
            Request:     wamp.GlobalID(),
            Topic:       testaments[i].topic,
            Arguments:   testaments[i].args,
            ArgumentsKw: testaments[i].kwargs,
            Options:     testaments[i].options,
        })
    }
}
sendTestaments(testaments.detached)
sendTestaments(testaments.destroyed)
```

aat/sessiontestament_meta_test.go
```diff
@@ -0,0 +1,146 @@
...
+      t.Fatal("Got event, shouldn't have event:", err)
+    }
+  case <-time.After(200 * time.Millisecond):
+  }
+
+  err = subscriber.Close()
+  if err != nil {
+    t.Fatal("Failed to disconnect subscriber:", err)
+  }
+}
```
gammazero: Please run `go fmt` on this file.

aat/sessiontestament_meta_test.go
```diff
@@ -0,0 +1,146 @@
...
+    if len(args) == 0 {
+      errChan <- errors.New("missing argument")
+      return
+    }
+    _, ok := wamp.AsString(args[0])
+    if !ok {
+      errChan <- errors.New("Argument was not string")
+      return
+    }
+    errChan <- nil
```
gammazero: Please have your event handler also check that the correct Arguments and ArgumentsKw values were received.

router/broker.go
```diff
@@ -28,6 +29,7 @@ var brokerRole = wamp.Dict{
 		featurePubExclusion:         true,
 		featurePubIdent:             true,
 		featureSubMetaAPI:           true,
+		featureTestamentMetaAPI:     true,
```
gammazero: According to the protocol specification, the feature is supposed to be announced by the _Dealer_ role:
https://wamp-proto.org/static/rfc/draft-oberstet-hybi-crossbar-wamp.html#feature-announcement-17

So, just move this in `dealer.go`

martin31821: I actually think this is a bug within the specification. Maybe @oberstet can help us here. 
Testament Meta API is listed within Advanced PubSub features, so I inferred to announce it within the broker. I will check against crossbario/crossbar reference implementation.

gammazero: Crossbar is compliant with the specification, and has `testament_meta_api` as a _Dealer_ feature:
https://github.com/crossbario/crossbar/blob/master/crossbar/router/dealer.py#L157

gammazero: To me it makes sense that is this a _Dealer_ feature since support or lack of does not affect what a client does with regard to the _Broker_, but it does affect what a client can do with regard to the _Dealer_, since this feature that is enabled or disabled is a meta-procedure.

martin31821: Changed it now, makes sense.

PR #110, opened 2018-06-15
Be more strict about what gets put into session meta details

gammazero:
Do not put details, other than those called for in the WAMP spec, into the on_join event and session_get response.  Doing so is not compliant with WAMP and may inadvertently expose private material stored in session details.

If additional details do need to be exposed, then these can be requested in the `RealmConfig`.

This change does cause a small behavioral change, and could break code that is depending on additional information being in the session meta details.   However, the change really is necessary IMO to protect against information exposure to session meta clients.  Any breakage is easily fixable by requesting the needed details fields.

### DIFF
--- a/aat/main_test.go
+++ b/aat/main_test.go
@@ -126,7 +126,8 @@ func TestMain(m *testing.M) {
 				RequireLocalAuth:  true,
 				RequireLocalAuthz: true,
 
-				ExtraSessionMetaDetails: []string{"foobar"},
+				MetaStrict:                true,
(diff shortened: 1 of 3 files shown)


ajgillis:
This latest change provides optional `RealmConfig.MetaStrict` flag to opt-in to restricting the session details provided by meta API.  By default, this is off and behavior is unchanged from previous.

gammazero:
This latest change provides optional `RealmConfig.MetaStrict` flag to opt-in to restricting the session details provided by meta API.

PR #118, opened 2018-06-27
Fix race condition with session testaments

gammazero:
When a session ends the session goroutine deletes the testaments for the session.  These testaments are stored in a map inthe realm.  This means that when multiple sessions end, their goroutines will concurrently modify the map causing a race condition.  This fixes the race by only allowing the realm's action goroutine to add or remove entries from the testaments map.


gammazero:
@martin31821 Please note change.

PR #121, opened 2018-07-01
Messagepack extensions

martin31821:
This PR simplifies the usage of handles of the codec library, since they are safe for concurrent reads by multiple encoders/decoders. This reduces overhead and memory consumption.

Furthermore I implemented a custom serialization for the BinaryData type which is not yet covered by the spec, but is required to recognize the intention of the user (send data as binary) in the client AND in the router. 
Therefore it is now possible to have reencoded messages in the router to keep the binary type.

### DIFF
--- a/transport/serialize/cborserializer.go
+++ b/transport/serialize/cborserializer.go
@@ -8,6 +8,13 @@ import (
 	"github.com/ugorji/go/codec"
 )
 
+var ch *codec.CborHandle
+
+func init() {
+	ch = &codec.CborHandle{}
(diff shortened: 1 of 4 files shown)


martin31821:
I'll take a look into reimplementing this in a more extensible way, i.e. to expose the serialization API to register custom types, since the original implementation for msgpack is very hacky.

I'd like to keep this as simple as possible to keep us within the spec.

martin31821:
I added some tests and a clean api.

PR #122, opened 2018-07-02
Add RPC benchmark test

ronanj:
The intention for this commit is to understand the RPC performance when using "large" argument structures.


aat/benchmarkrpc_test.go
```diff
@@ -0,0 +1,143 @@
...
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+
+		result, err := client.Call(ctx, "action", nil, callArgs, nil, "")
+		if err != nil {
+			panic(err)
+		}
+		if i < 2 {
+			verify(result)
```
gammazero: Even though the number of calls to `verify()` is limited, you may want to consider calling it outside of the benchmark loop, after calling `b.StopTimer()`

gammazero: A note about benchmarking nexus in general: The nexus router is designed to be highly concurrent.  It tries to handle as much as possible on concurrent goroutines, except for the aspects or routing that need access to shared resources.

What this means is that a benchmark for traffic between any two clients will not be better than a  totally serialized/single-threaded router.  The performance advantage of nexus should be seen as the number of clients increases.  A benchmark that ramps up the number of clients would also be good.

PR #129, opened 2018-07-11
Allow another publish filter to be used

martin31821:
This exposes the publish filter interfaces and structs to allow embedding applications to define additional filter mechanisms.

No breaking API change here, the behavior is the same as before.

### DIFF
--- a/router/broker.go
+++ b/router/broker.go
@@ -461,16 +461,15 @@ func allowPublish(sub *session, filter PublishFilter) bool {
 	if filter == nil {
 		return true
 	}
-	if filter.LockRequired() {
-		sub.RLock()
-		defer sub.RUnlock()
-	}
 	// Create a safe session to prevent access to the session.Peer.
 	safeSession := wamp.Session{
 		ID:      sub.ID,
 		Details: sub.Details,
 	}
-	return filter.PublishAllowed(&safeSession)
+	sub.rLock()
+	ok := filter.Allowed(&safeSession)
+	sub.rUnlock()
+	return ok
 }
 
 // pubEvent sends an event to all subscribers that are not excluded from
@@ -615,13 +614,13 @@ func disclosePublisher(pub *session, details wamp.Dict) {
 	details[rolePub] = pub.ID
 	// These values are not required by the specification, but are here for
 	// compatibility with Crossbar.
-	pub.RLock()
+	pub.rLock()
 	for _, f := range []string{"authid", "authrole"} {
 		if val, ok := pub.Details[f]; ok {
 			details[fmt.Sprintf("%s_%s", rolePub, f)] = val
 		}
 	}
-	pub.RUnlock()
+	pub.rUnlock()
 }
 
 // ----- Subscription Meta Procedure Handlers -----
--- a/router/broker.go
+++ b/router/broker.go
@@ -114,10 +114,14 @@ func (b *Broker) Publish(pub *session, msg *wamp.Publish) {
 	if pub == nil || msg == nil {
 		panic("broker.Publish with nil session or message")
 	}
+	// Send a publish error only when pubAck is set.
+	pubAck, _ := msg.Options[wamp.OptAcknowledge].(bool)
+
 	// Validate URI.  For PUBLISH, must be valid URI (either strict or loose),
 	// and all URI components must be non-empty.
+
 	if !msg.Topic.ValidURI(b.strictURI, "") {
-		if pubAck, _ := msg.Options[wamp.OptAcknowledge].(bool); !pubAck {
+		if !pubAck {
 			return
 		}
 		errMsg := fmt.Sprintf(
@@ -145,12 +149,17 @@ func (b *Broker) Publish(pub *session, msg *wamp.Publish) {
 	if opt, _ := msg.Options[wamp.OptDiscloseMe].(bool); opt {
 		// Broker MAY deny a publisher's request to disclose its identity.
 		if !b.allowDisclose {
-			b.trySend(pub, &wamp.Error{
-				Type:    msg.MessageType(),
-				Request: msg.Request,
-				Details: wamp.Dict{},
-				Error:   wamp.ErrOptionDisallowedDiscloseMe,
-			})
+			if pubAck {
+				b.trySend(pub, &wamp.Error{
+					Type:    msg.MessageType(),
+					Request: msg.Request,
+					Details: wamp.Dict{},
+					Error:   wamp.ErrOptionDisallowedDiscloseMe,
+				})
+			}
+			// When the publisher requested disclosure, but it isn't
+			// allowed, don't continue to publish the message.
+			return
 		}
 		disclose = true
 	}
@@ -164,7 +173,7 @@ func (b *Broker) Publish(pub *session, msg *wamp.Publish) {
 	}
 
 	// Send PUBLISHED message if acknowledge is present and true.
-	if pubAck, _ := msg.Options[wamp.OptAcknowledge].(bool); pubAck {
+	if pubAck {
 		b.trySend(pub, &wamp.Published{Request: msg.Request, Publication: pubID})
 	}
 }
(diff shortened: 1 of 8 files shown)


router/broker.go
```diff
@@ -147,13 +154,14 @@ func (b *Broker) Publish(pub *session, msg *wamp.Publish) {
...
 			})
+			return
 		}
 		disclose = true
 	}
 	pubID := wamp.GlobalID()
 
 	// Get blacklists and whitelists, if any, from publish message.
-	filter := newPublishFilter(msg)
+	filter := b.filterFactory(msg)
```
ajgillis: This step was done for the particular filter implementation I was using, to "pre-compile" the block-whitelists into maps.  I am not sure this is the right approach for letting users provide their own filters.   I need to think about this.  

If this is the right approach then need to document/encourage implementer to do as much much work as possible in the FilterFactory function, and not in the filter itself (since the filter is executed in the broker's critical section).

In general, I want to avoid making it possible for the user to define something that is executed in a broker or dealer critical section (single core goroutine that accesses mutable state).  If the user needs to supply some functionality, it should be runable in the session's message handling goroutine, not in core of broker or dealer.  Since matching against subscriber must be done in core, every effort must be taken to minimize what is done there.  I think many users trying to use the feature will end up slowing down their router and not know why.

**An Alternative Approach**
An alternative approach is to create a client-side filter tool.  This uses the session meta API to get information about all sessions, and to keep that info updated.  When the client wants to publish a message, it runs its client filter and selects a list of session IDs it wants to publish to, it then puts these IDs into a simple whitelist and publishes the message.

Consider the performance advantage of the above.  Suppose there are 500 clients trying to publish messages to a filtered set of subscribers .  The client-side tool would distribute most of the filter processing over the clients.  Also, the filtering could be as complex and slow as needed -- consider looking up security policy and ACLs in a database to decide who to publish to.

router/publishfilter.go
```diff
@@ -79,17 +85,17 @@ func newPublishFilter(msg *wamp.Publish) *publishFilter {
...
-// publishAllowed determines if a message is allowed to be published to a
+// PublishAllowed determines if a message is allowed to be published to a
 // subscriber, by looking at any blacklists and whitelists provided with the
 // publish message.
 //
 // To receive a published event, the subscriber session must not have any
 // values that appear in a blacklist, and must have a value from each
 // whitelist.
-func (f *publishFilter) publishAllowed(sub *session) bool {
+func (f *simplePublishFilter) PublishAllowed(sub *Session) bool {
```
ajgillis: I do not want to expose the `router.session`.  This is a wrapper around `wamp.Session` and contains items that should remain private to the router.  If you have `PublishAllow()` take a `*wamp.Session` and move the locking outside of the call to `PublishAllowed()`, I think that will accomplish.

I really want to make every effort to not force the user to decide when to read-lock or write-lock a session.  If we make the user do that, then we need to reexamine the design. 

Again, consider implementing client-side filtering.

martin31821: That's right, didn't consider to move the locking out of `PublishAllowed`, will change this one.

martin31821:
Overall I don't see a client-side filtering better than server-side, because the client has (for each publication, if done right) the list of sessions, call `wamp.session.get` for each session and perform its logic afterwards. I think that would put actually more load on the server than having only server-side filtering.

gammazero:
@martin31821 The idea was that the client would only call `wamp.session.get` once at the beginning of the its session, and rely on the subscription meta event to keep the list of subscribers up-to-date.

martin31821:
I'd also like to see a client-side filtering implemented, would you like to see this as part of nexus or as standalone library? 

I'd prefer a standalone library, which can be initialized from a nexus client instance.

gammazero:
I have not forgotten about this PR, just need some more time to think about it.

I have a client side implementation that I will brush the dust of off and give to you to see what you think. I agree that it would be best as a separate library.  It is really a tool for client developers, and not part of the client itself.

martin31821:
It's ok, I used the API to create composable filters in autobahnkreuz as extension, maybe you want to take a look (this was the use case I originally intended to implement in nexus, but I thought since it's not within the spec, it shouldn't be here, my use case was that I wanted to publish an event to all members of an authrole AND a specific authid, which isn't possible using the upstream WAMP publish filter).  ([Reference](https://github.com/EmbeddedEnterprises/autobahnkreuz/tree/master/filter))

gammazero:
What have is a client side monitor, that monitors sessions coming and going using session meta events.  You can add filters to the monitor, a filter is just a user-defined function that returns `true` or `false`.  When a new session joins, the monitor receives the wamp.session.on_join message and calls the filter with the session details.  If the filter function returns true, the session is added to a whitelist.

At any time a user can retrieve the whitelist associated with any filter.  The whitelist is always up-to-date.

gammazero:
@martin31821 Here is what I had in mind for the client filtering.  It is very flexible since arbitrarily complex filtering functionality can be defined without affecting router performance, since filter whitelist is calculated on client-side.

https://github.com/gammazero/nexus/blob/filtertool/client/filtertool/filtertool.go

martin31821:
That seems like a fairly good starting point. I think it would be required to react to changes made via the session edit meta procedure.

I will evaluate this and check whether it works for our use case (weak cpu, around 400mhz).

gammazero:
@martin31821 The filtertool is a stand-alone package, in that nothing in nexus uses it and it would be up to a client implementer to make use of it.  However, if this something you want to build on or include as part of a larger set of tools (or just want to write a better version of), then I would be happy to move it into a separate repo where you can do develop freely, instead of maintaining this within the nexus repo.

Than may be better if you want to use that as a starting

martin31821:
@gammazero I finished evaluating, but came to the conclusion that it's not applicable for our use case.

To give some context:
We have a few senders connected over wireless network to the router, which send messages frequently to (potentially large) groups of connections attachted to the router via internet.
Large groups means 1k-50k connections, which results in two problems for us:

1. each message from the sender to the router includes the mentioned 1-50k connection IDs, which blows the message size up (msgpack is like 400kB data for 50k connections, JSON like 1.5-2MB)
2. We get even more traffic on the bandwidth-restriced wireless network due to the join/leave messages and following session queries to the server.

gammazero:
@martin31821 Thank you very much for your analysis and feedback.  This is a very good example: In your case, the client-side filtering does not work well since every published message will need to have a very large number of IDs, sent over a slow network.   Communication time is the limiting factor, not the time to do the filtering.

I will reevaluate your PR since it is clear there are cases where user-defined filter logic may be necessary.

gammazero:
@martin31821 After additional review, I want to merge this PR (finally).  First, I have a question about `session.RLock` and `session.RUnlock`:  Do these need to be exported?  With the `filter.LockRequired()` function is needed anymore?
https://github.com/gammazero/nexus/pull/129/files#diff-95aef7620f08580475d2dfa823437d56R465

I would also propose removing `filter.LockRequired()` and always acquiring the `RLock`.

martin31821:
> @martin31821 After additional review, I want to merge this PR (finally). First, I have a question about `session.RLock` and `session.RUnlock`: Do these need to be exported? With the `filter.LockRequired()` function is needed anymore?
> https://github.com/gammazero/nexus/pull/129/files#diff-95aef7620f08580475d2dfa823437d56R465

That's correct, I initially thought about allowing the implementer to perform locking itself, but then moved away from this approach for obvious reasons.

> I would also propose removing `filter.LockRequired()` and always acquiring the `RLock`.

I'm unsure about this tbh, but I can't think of any case where obtaining the RLock does any harm. Mostly because sessions RWLock should occur infrequently during execution time, so RLock should be a cheap operation.
I agree on documenting the change and marking it as unsafe/performance critical.

gammazero:
@martin31821 If you are good with my changes to this PR then I will merge it.

martin31821:
LGTM :+1: 
This will finally allow us to build our router against upstream

PR #130, opened 2018-07-13
Dealer cancels progressive results if caller leaves realm

gammazero:
- Send INTERRUPT to callee to stop progressive results if caller left realm
- Disable progressive call results feature of callees that do not support call cancelling
- Cleanup pending calls and invocations when client leaves realm
- Additional arguments in wamp.session.on_leave event
- Test interrupting progressive results in the middle by closing caller
- Cleanup invocation handlers if client closes during RPC handling
- Return error to call if caller closed while waiting on call result

### DIFF
--- a/aat/rpcprogress_test.go
+++ b/aat/rpcprogress_test.go
@@ -33,19 +33,16 @@ func TestRPCProgressiveCallResults(t *testing.T) {
 		if e != nil {
 			fmt.Println("Error sending Alpha progress:", e)
 		}
-		time.Sleep(500 * time.Millisecond)
 
 		e = callee.SendProgress(ctx, wamp.List{"Bravo"}, nil)
 		if e != nil {
 			fmt.Println("Error sending Bravo progress:", e)
 		}
-		time.Sleep(500 * time.Millisecond)
 
 		e = callee.SendProgress(ctx, wamp.List{"Charlie"}, nil)
 		if e != nil {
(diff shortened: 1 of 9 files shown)


gammazero:
Here is a sequence of events that describes how progressive results are handled when the caller leaves in the middle of receiving them:

1. Caller sends call, willing to receive progressive results
2. Callee starts sending progressive results
3. Caller receives progressive results
4. Caller closes/exits realm
5. Dealer removes caller session and any pending calls
6. Caller (user code) gets "client closed" error from `CallProgress()` if call was not already canceled. 
7. Callee (user code) sends another progressive result
8. Dealer sends INTERRUPT to callee to cancel call
9. Callee (user code) sends another progressive result (before handling INTERRUPT)
10. Dealer sends another INTERRUPT to the callee
11. Callee (nexus code) receives first INTERRUPT and cancels the call
12. Callee (nexus code) receives second INTERRUPT for call that does not exist, discards
13. Callee (user code) tries to send another proressive result, and gets error that caller is not accepting progressive results. 

_note: well-behaved callee user code should exit its invocation handler at this point due to call context being canceled or due to error from_ `SendProgress()`

14. Callee (user code) ignores errors and call context, and returns final result.  Final result goes nowhere since call is canceled and nothing is processing the result.

gammazero:
@HolloTooth Try this out

Issue #132, opened 2018-07-18
wamp subscription api procedures missing


dionytadema:
according to the documentation nexus should support the wamp subscription api, but it seems like the procedures are missing

are there any plans to implement those? they would really help me out

gammazero:
You can expect these soon, within a few days most likely.

The documentation needs to break `subscription_meta_api` into `subscription_meta_events` and `subscription_meta_procedures`, the latter of which are not there at the moment.  The subscription meta procedures are a newer addition to the spec, so at some point in the past the documentation was correct, but had since become inaccurate.  Will fix that too.

martin31821:
Don't expect the procedures to be that easy to add. I had a try implementing these myself, but I failed since the procedure description is very vague and (I think) tied to a specific implementation architecture, which is, if I understood correctly, not the subscription scheme used within nexus...

gammazero:
@martin31821 , I am almost done.  The reason it was difficult was because the map structure in the broker needed to change somewhat.

dionytadema:
Wow, thanks for the quick reply.

You are doing great work and nexus is by far one of the best routers for wamp!

gammazero:
Fixed by #134 which is merged. During stress testing a critical client issue was discovered and also fixed.

@martin31821 The broker has been restructured and is much cleaner (and correct now).  Once the map structure was corrected, implementing the meta procedures was straight forward.  Please update to get some important fixes.  If you are considering further modifications to the code, I think you will find the new broker much nicer to work with.

PR #134, opened 2018-07-20
Implement subscription meta procedures

gammazero:
This implements the following subscription meta procedures:
- wamp.subscription.list
- wamp.subscription.lookup
- wamp.subscription.match
- wamp.subscription.get
- wamp.subscription.list_subscribers
- wamp.subscription.count_suscribers

This also cleans up how the broker tracks subscriptions and subscribers.


gammazero:
Fixes #132 

Stress testing uncovered a critical issue: possibility to deadlock client.  This has been fixed so that the cause of the deadlock has been removed; is no longer possible.

Issue #135, opened 2018-07-20
Router panics with KeepAlive


HollowTooth:
Hi, 
I just recently added wss to the nexus router, and noticed, that the router panics, if `KeepAlive` is set (not 0).
This is related to one of my older issues here #99.
The problem, that I found, is that you cast the `net.Conn` to `net.TCPConn`, when with wss you have `*net.TLSConn`.
````
if s.KeepAlive != 0 {
	conn.UnderlyingConn().(*net.TCPConn).SetKeepAlive(true)
	conn.UnderlyingConn().(*net.TCPConn).SetKeepAlivePeriod(s.KeepAlive)	
}
````

gammazero:
I will fix this within 12 hours (as soon as I have a moment this evening).  Will use websocket ping/pong as keepalive.

gammazero:
@HollowTooth Fix is now merged.  A couple of other important fixes are included with latest merge, please update to latest version of nexus.

Thank you for reporting this problem.  It lead to a better way to handle the problem, as well as much more testing and improvements.

PR #136, opened 2018-07-21
Use websocker ping/pong as keepalive for websockets.

gammazero:
The WebsockerServer.KeepAlive configures a websocket "ping/pong" heartbeat when set to a non-zero value, where the value of KeepAlive is the interval between heartbeats.  If a "pong" response is not received after 2 intervals have elapsed then the websocket connection is closed.

This fixes issue #135

### DIFF
--- a/aat/main_test.go
+++ b/aat/main_test.go
@@ -161,6 +161,7 @@ func TestMain(m *testing.M) {
 		}
 		s.EnableTrackingCookie = true
 		s.EnableRequestCapture = true
+		s.KeepAlive = time.Second
 		closer, err = s.ListenAndServe(tcpAddr)
 	case "https", "wss":
 		s := router.NewWebsocketServer(nxr)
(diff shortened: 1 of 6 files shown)


dionytadema:
this seems to break compatibility with autobahn pyton, i get a 
`peer not responging to pings, closing websocket`
error in the router while my python programs is publishing some data in a loop. do you have any idea why this might occur

gammazero:
I tested with the autobahn python clients in [examples/interop](https://github.com/gammazero/nexus/tree/master/examples/interop/autobahn-python) and it is able to send ping pings to the router and receive pongs.  It also receives pings from the router and responds with pongs.  So, that particular autobahn client is working.

If you are publishing data in a loop, and using the twisted python client, then it sounds like the problem is that your loop needs to yield control to the reactor occasionally so that the reactor can handle the incoming pings.  The way to do this is to insert the following into your loop:
```
yield deferLater(reactor, 1, lambda: None)
```
That will yield control to the reactor for a short amount of time to let it deal with the ping/pong handling, and then resume your loop.

You may also need to include the following:
```
from twisted.internet import reactor
from twisted.internet.task import deferLater
```

PR #137, opened 2018-07-27
Fix prefix/wildcard invocation handling

martin31821:
A prefix registration was ignored when the prefix was empty.
Additionally, the spec requires a router to prefer prefix matches over wildcard matches and send the actual procedure the client tried to call as `procedure` field in the details dict.

### DIFF
--- a/router/dealer.go
+++ b/router/dealer.go
@@ -627,8 +627,7 @@ func (d *Dealer) call(caller *session, msg *wamp.Call) {
 					Details: wamp.Dict{},
 					Error:   wamp.ErrOptionDisallowedDiscloseMe,
 				})
-			}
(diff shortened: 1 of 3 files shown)


router/dealer.go
```diff
@@ -516,7 +516,7 @@ func (d *Dealer) matchProcedure(procedure wamp.URI) (*registration, bool) {
 		// No exact match was found.  So, search for a prefix or wildcard
 		// match, and prefer the most specific math (longest matched pattern).
 		// If there is a tie, then prefer the first longest prefix.
-		var matchCount int
+		matchCount := -1 // initialize matchCount to -1 to catch an empty registration.
```
gammazero: Nice!

Issue #138, opened 2018-07-29
Websocket: Possibility to change the WS write channel length


ronanj:
in `websocketpeer.go`, the following constants are used:

```go
const (
	outQueueSize = 16
)
```

Is it possible to provide a way to modify this `outQueueSize` value?

gammazero:
With the latest version of nexus, if the write channel is blocked, then the router will wait for a second to see if the message can be sent before dropping it.  So, if that ever happens, it means that the underling IPC is failing to buffer additional data.  Not sure that increasing the size would really help much.

I think the most technically correct thing to do is to set the size to 1 and rely on the short delay to wait for the kernel to buffer the waiting messages.  On a severely overloaded machine, it might be possible that the kernel is not able to buffer additional messages in time, but if this happens for more than a few messages, then something is wrong.  Having a channel size of 16 seems more than generous.  The situation may be slightly different when the peer is a local peer and does not have an underlying IPC mechanism (socket).

My concern here is that increasing the channel size will only mask the cause of the real issue.  If you have a particular case where messages are getting dropped, I would like to look into this to determine if changing the channel size really is the solution, or if this is the result of a larger issue.

ronanj:
I agree on that increasing the size will only mask the real issue, so if the real issue can be solved then that would definitely be even better.

The example I am using is available in this commit: https://github.com/ronanj/nexus/commit/12dbf085cecea1ee6ad0d0fcaea950e0a5b95e12 

It consists of a NodeJs callee and Golang router/caller - where the callee tries to send as many messages as fast as possible. The error `!!! dealer dropped RESULT message: blocked` happens systematically when the callee sends the response back to the caller.

gammazero:
I see that the version of nexus that you based your fork on does not have the fix for this issue, that I mention above.  You need to get the most recent version.  See if that fixes what you are seeing.

ronanj:
Fixed with version 2.1.0. Thanks

PR #139, opened 2018-07-30
Add nodejs/go interop example

ronanj:
This example is related to issue https://github.com/gammazero/nexus/issues/138

### DIFF
--- a/examples/interop/wampy-nodejs/callee.go
+++ b/examples/interop/wampy-nodejs/callee.go
@@ -1,0 +1,106 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"encoding/json"
+	"github.com/gammazero/nexus/client"
+	"github.com/gammazero/nexus/router"
+	"github.com/gammazero/nexus/wamp"
+	"io"
+	"log"
+	"os"
+	"os/signal"
+	"time"
+)
+
+const procedureName = "server.time"
+const wwsPort = 8008
+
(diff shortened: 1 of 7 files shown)


examples/interop/wampy-nodejs/callee.go
```diff
@@ -0,0 +1,106 @@
...
+	"os"
+	"os/signal"
+	"time"
+)
+
+const procedureName = "server.time"
+const wwsPort = 8008
+
+func main() {
+	broker := newBroker(wwsPort)
```
ajgillis: `Broker` is probably not a good name here.  WAMP defines a `broker` role with functionality to handle pub/sub.  Calling you RPC server `broker` may confuse people who think your broker has something to do with the WAMP broker.

May I suggest calling this `timeServer`

examples/interop/wampy-nodejs/callee.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// Register procedure "sum"
```
ajgillis: Comment is wrong.  It is registering `"server.time"`.

examples/interop/wampy-nodejs/callee.go
ronanj: Agree on the confusion on the word `broker` (sorry for the novice mistake). I have updated the example, however, used the name `WampClient`

PR #163, opened 2018-12-21
Fix misaligned call routing, use session scoped IDs within the client

martin31821:
When two clients are connecting concurrently and performing long running calls with the same request ID,
the router can reassign call results to the wrong session, which may lead to a DoS and/or information leaks
since protected call results can be rerouted to not authorized callers.

Furthermore, a client can 'steal' the result of another client, causing its call to block for
infinite time, maybe causing a DoS.

@gammazero

/cc @dionytadema please test this against your issue, I wasn't able to reproduce this anymore with this changes.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -156,6 +156,7 @@ type Client struct {
 	closed        int32
 	done          chan struct{}
 	routerGoodbye *wamp.Goodbye
+	idGen *wamp.SyncIDGen
 }
 
 // NewClient takes a connected Peer, joins the realm specified in cfg, and if
@@ -199,6 +200,7 @@ func NewClient(p wamp.Peer, cfg Config) (*Client, error) {
 
 		log:   cfg.Logger,
 		debug: cfg.Debug,
+		idGen: new(wamp.SyncIDGen),
 	}
 	go c.run() // start the core goroutine
 	return c, nil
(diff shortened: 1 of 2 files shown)


martin31821:
FYI: The build failure seems to be related to travis-ci, I tested locally against `golang:rc` docker container (1.12beta1) and it worked fine.

dionytadema:
I tested this fix against the example i made and our own codebase and it seems to have fixed the issue

thanks very much for the fast fix!

gammazero:
The code in this PR looks good, but I want to make sure I understand the problem completely before merging.

How is it that two concurrent clients are ever using the same request ID?  The request IDs are large random numbers so the chance of collision or guessing a request should be very low.  That would seem like an error in the implementation of the global ID generator.

martin31821:
The spec defines request IDs as session scope (= up-counting from one for
each client).

The nexus client implementation didn't use session scope, but global IDs as
request ID, which is why it is working against nexus.



Am Di., 25. Dez. 2018, 16:08 hat Andrew J. Gillis <notifications@github.com>
geschrieben:

gammazero:
@martin31821 I see that requests are session scope in this section:
https://wamp-proto.org/_static/gen/wamp_latest.html#ids

But for `call` the spec states that the ID is random:
https://wamp-proto.org/_static/gen/wamp_latest.html#call-0

That is the source of the confusion.  Looks like the spec need a bit more cleanup.

gammazero:
Will merge.  FYI:
https://github.com/wamp-proto/wamp-proto/issues/270

Issue #159, opened 2018-11-14
Router is blocked for several minutes after message bursts


martin31821:
I'm running into problems with the sendTimeout when clients are blocked, I try to describe my problem:

- I have some clients (like 300), some of them have slow network connections and are subscribed to 'hot' topics (frequently published to)
- The router is correctly blocked and drops some messages according to https://github.com/gammazero/nexus/blob/master/transport/websocketpeer.go#L165
- Eventually, some clients get dropped (because the pingpong mechanism fails)
- The messages from the dropped clients are processed **after** the client is gone, causing the router to block for several minutes.
- This in turn blocks the router from adding and processing new sessions.

The problem I identified is that the 'session died' handler is processed synchronously with the pending requests for all sessions, but I'm not sure what's the best way to improve the situation.

I also have a repro [here](https://github.com/EmbeddedEnterprises/nxstest)

martin31821:
After some hours of debugging this, I now have another suspiction. I saw that registrations are still present within the router state even if the connections are not alive anymore, which I think happens iff I run into `peer is not responging to pings, closing websocket` which is not passed through correctly to the upper layers.

gammazero:
Do you know if this is a new behavior introduced by the recent update of gorilla websockets (#157)?

I will start working on this now.

martin31821:
This is unrelated to the vendor updates, we experienced the problem in a production setup running an older version (#155)

One problem I see is that the router is blocked when waiting for clients to become available, which is a great problem, since the incoming messages will queue up indefinitely (push to actionChan in broker and dealer) and while queueing up, lose their order, which leads eventually to registrations or subscriptions which have been established for an already closed connection. 

I tried to debug it [here](https://github.com/EmbeddedEnterprises/nexus/tree/feature/debug), adding more logging and tried to fix it, but that didn't work either.

gammazero:
I will have this fixed ASAP.

gammazero:
The main cause of the problem is when the output channel for a slow client is full.  This causes a temporary block waiting to write the  channel, which blocks in the critical section of the broker.  When there are a number of such clients, this becomes particularly bad.  The original intent was to give client output goroutine time to read the outgoing messages from the channel and write them to the socket.  This fails badly when the socket is blocked and the message on the output channel cannot be removed.

[...]

PR #167, opened 2019-03-14
Fix blocking with slow clients

gammazero:
This fixes issue #159.  When a client leaves the realm or is disconnected by the router, do not attempt to send more messages to the client.

Web and raw socker server's outbound per-client channel size is configurable.

Progressive results use backpressure to limit influx of results when results are being produced faster than they are being consumed.


gammazero:
I am not comfortable with allowing the server's outbound per-client channel size to be configurable, because it is a channel, not a queue.  The typical case for resizing would be to provide more buffer for slow clients receiving message bursts.  I have an implementation where the channel is replaced by a dynamic queue (actually queue between a read and a write channel), but that adds complexity and other risks.

stevecalvert:
Hi @gammazero, just wondering if you were planning any further activity on this PR.  We think we are seeing similar behaviour as reported by #159, as evidenced by a steep rise in memory allocation (we also saw the goroutines used by the wamp router go through the roof before the process was killed off due to OOM by the kernel).  We're using build commit #f41714b.  We had 1500 client websockets established (all RPC, no pub-sub) taking around 5000 goroutines.  At the point of failure the go-routine count rose in excess of 25k before termination.  At the same time, allocated memory went from a steady 100MB to last recorded 2.3GB before OOM

gammazero:
@stevecalvert Yes, this has been sitting too long and needs this fix is needed!  I was holding off until I could get more testing done.  I plan to do that and merge this weekend.  If there is anyway you could see if this branch fixes what you are seeing (or if not), that would be helpful feedback.

gammazero:
@martin31821 Finally had time to test and complete this PR.  This should handle the problems reported in #159.

Additionally, this will handle the case where a callee is producing progressive results faster than the caller can consume them.  That is likely to be a typical situation when a caller requests a large body of data, particularly if that data requires some time for the caller to process or store, and the data is delivered in chunks via progressive results.

martin31821:
Great news. When you write about progressive results: what happens during the situation you mention above?

gammazero:
@martin31821 Before PR, what happens with progressive results arriving from the callee faster than they can be sent to the callee, is they fill the callee's send queue (channel) until it is blocked.  Once that happens, the dealer is blocked while each response waits for a `sendTimeout` before it is sent or dropped:
https://github.com/gammazero/nexus/blob/master/transport/websocketpeer.go#L178

This is a problem either way, because the whole dealer is blocked waiting for each progressive result  to be dropped or queued.  So all communication through the dealer are delayed, not only the progressive results -- and this happens with each progressive result.  This is similar to the problem you are seeing with the slow clients.  Also, if the result is dropped because the caller is blocked for too long, then the caller is missing a progressive result.  This is unacceptable as the overall result will be incomplete.  If it is ok to discard a result, then only the caller can make that decision.

This PR changes the behavior when the callee sends progressive results faster than the caller can handle them.  Now the dealer is not blocked.  Instead, only the callee waits and retries sending the result until a (much longer) timeout, or until the caller leaves. This callee wait-retry causes back-pressure on the callee, preventing callee from sending more progressive results until there is room on the caller's outbound queue (without blocking dealer).  If the caller does not make any room on queue for a long timeout, then the whole call is canceled.

martin31821:
That seems fairly reasonable, I really like the idea on putting back pressure on the callee. 
I'm currently facing the same problem with progressive call results and our decorator tests.

Really excited to see this fix merged :fire:

PR #169, opened 2019-03-28
Feature/improve rpc benchmark

bkneis:
Currently the RPC benchmark tests creates a single client and sends a large number of requests to the server. I think a more realistic test is to create a large number of clients that send a number of requests to the server.

Additionally the http and https scheme for the benchmark did not use the serveHTTP handler to upgrade a http connection to a websocket. I thought it would be good to allow these options to perform this so that you can see the overhead of using http/https instead of direct ws/wss.

This PR addresses these 2 items, the first simply takes the code that creates the client and performs a request into a go routine and allows the main loop to spawn many of these routines. We then use a wait group and block the test until all clients are done.

The second was to add an explicit case for http and https in the main_test to serve a http server that can handle the serveHTTP on an endpoint used by connectClientCfg()


gammazero:
I agree that a benchmark that scales up the number of clients is desirable.

The benchmark that was there before this change scaled up only the number of requests.  The intent was to see how code changes affected end-to-end message handling speed, with the assumption that more clients scaled linearly for any number of messages.

I think it would be most informative to have two types of benchmark:
1. Scale number of requests with one or fixed number of clients.
2. Scale number of clients, each with a fixed number of requests

A potential problem with your benchmark is if the time to establish the HTTPS connection + wamp session is large compared to the time to send a message, then what you are actually benchmarking is wamp session setup time, not message throughput.

Consider if the original benchmark is used, but have a series of benchmarks that do it with 2 clients, then 4 clients, then 8, then 16 until some 2^N clients.  Then we see how message throughput scales with different numbers of clients.

bkneis:
@gammazero Agree with everything you said. I'll update the PR so that the benchmark tests the performance in relation to dynamic number of requests and also clients as separate tests.

I understand for the https connection this would include set up time, but this is what we intentionally wanted to do. We are currently using nexus in production and running our performance tests on our software. What we would like to know if from an end to end perspective how long it would take for N clients to connect and send a message. It wasn't really to actually benchmark the RPC call at all. With this in mind maybe we can add a https-with-setup? Or set a flag or something similar as this would be very useful to us.

aat/benchmark_rpc_test.go
```diff
@@ -148,27 +148,38 @@ func benchmarkRpc(b *testing.B, action client.InvocationHandler, callArgs wamp.L
...
+			var result *wamp.Result
+
+			for j:= 0; j < 10; j++ {
+				result, err = client.Call(ctx, "action", nil, callArgs, nil, "")
+				if err != nil {
+					panic(err)
+				}
+				verify(result)
+			}
+		}()
```
gammazero: When looking at this... it creates and connects N clients, each doing 10 RPC calls...  I was not sure if it was trying to test how fast clients could be connected, or testing how fast RPC calls were being made.  May I suggest making a new benchmark like this, except it only connects N clients - no RCP. - Probably makes sense just to do that with HTTP/HTTPS.

If I am missing an important reason for the RPC calls, please let me know.  Otherwise, it just looks like it conflates what you are really trying to benchmark.

aat/main_test.go
```diff
@@ -278,10 +325,18 @@ func connectClientCfg(cfg client.Config) (*client.Client, error) {
 
 	var addr string
 	switch scheme {
-	case "http", "ws", "tcp":
-		addr = fmt.Sprintf("%s://%s/", scheme, tcpAddr)
+	case "http", "https":
```
gammazero: In this `switch`, `"http"` should be handled the same as `"ws"` and `"tcp"`.  `"https"` should be handled the same as `"wss"` and `"tcps"`

aat/main_test.go
```diff
@@ -157,7 +170,41 @@ func TestMain(m *testing.M) {
...
+			webRouter := mux.NewRouter()
+			webRouter.HandleFunc("/ws", s.ServeHTTP).Methods("GET")
+			wampServer := &http.Server{
+				Addr:    httpAddr,
+				Handler: webRouter,
+			}
+
+			// Check if we should use the certs when using tls
+			if scheme == "http" {
+				err = wampServer.ListenAndServe()
```
gammazero: Won't this cause a race condition with different goroutines writing to the same `err`?  Maybe assign to something other than `err`  in goroutine.

aat/main_test.go
```diff
@@ -157,7 +170,41 @@ func TestMain(m *testing.M) {
...
+			} else {
+				err = wampServer.ListenAndServeTLS(certPath, keyPath)
+			}
+			if err != nil {
+				fmt.Print("Error ", err.Error())
```
gammazero: Print to `os.Stderr`

aat/main_test.go
```diff
@@ -93,6 +96,16 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
+	fmt.Println("Increasing ulimit to max")
+	var rLimit syscall.Rlimit
+	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit); err != nil {
```
gammazero: declaration of "err" shadows declaration at line 46

aat/main_test.go
```diff
@@ -93,6 +96,16 @@ func TestMain(m *testing.M) {
...
+		panic(err)
+	}
+	rLimit.Cur = rLimit.Max
+	if err := syscall.Setrlimit(syscall.RLIMIT_NOFILE, &rLimit); err != nil {
```
gammazero: declaration of "err" shadows declaration at line 46

PR #170, opened 2019-04-01
Add timeout to Dial to avoid infinite timeout

bkneis:
Currently the websocket dialer uses the Dial function that internally calls DialContext with a background context. Since a background context does not contain a timeout, an infinite timeout is possible causing the call to block. This PR simply supplies a context with a 5 second timeout.

Additionally the Dial function used from the net package did not use a timeout, again I have added a 5 second timeout.

### DIFF
--- a/transport/rawsocketpeer.go
+++ b/transport/rawsocketpeer.go
@@ -84,9 +84,9 @@ func ConnectRawSocketPeerTimeout(
 	}
 
 	if timeout == 0 {
-		conn, err = net.DialTimeout(network, address, timeout)
-	} else {
 		conn, err = net.Dial(network, address)
+	} else {
+		conn, err = net.DialTimeout(network, address, timeout)
 	}
 
 	if err != nil {
--- a/transport/rawsocketpeer.go
+++ b/transport/rawsocketpeer.go
@@ -64,7 +64,7 @@ func ConnectRawSocketPeer(network, address string, serialization serialize.Seria
 		return nil, err
 	}
 
-	conn, err := net.Dial(network, address)
+	conn, err := net.DialTimeout(network, address, 5 * time.Second)
 	if err != nil {
 		return nil, err
 	}
--- a/transport/rawsocketpeer.go
+++ b/transport/rawsocketpeer.go
@@ -1,6 +1,7 @@
 package transport
 
 import (
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
@@ -45,21 +46,22 @@ const (
 	magic = 0x7f
 )
 
-// ConnectRawSocketPeer calls ConnectRawSocketPeerTimeout without a Dial timeout
+// ConnectRawSocketPeer calls ConnectRawSocketPeerContext without a Dial
+// timeout
 func ConnectRawSocketPeer(network, address string, serialization serialize.Serialization, logger stdlog.StdLog, recvLimit int) (wamp.Peer, error) {
-	return ConnectRawSocketPeerTimeout(0, network, address, serialization, logger, recvLimit)
+	return ConnectRawSocketPeerContext(context.Background(), network, address, serialization, logger, recvLimit)
 }
 
-// ConnectRawSocketPeerTimeout creates a new rawSocketPeer with the specified config,
-// and connects it to the WAMP router at the specified address.  The network
-// and address parameters are documented here:
+// ConnectRawSocketPeerContext creates a new rawSocketPeer with the specified
+// config, and connects it to the WAMP router at the specified address.  The
+// network and address parameters are documented here:
 // https://golang.org/pkg/net/#Dial
 //
 // If recvLimit is > 0, then the client will not receive messages with size
 // larger than the nearest power of 2 greater than or equal to recvLimit.  If
 // recvLimit is <= 0, then the default of 16M is used.
-func ConnectRawSocketPeerTimeout(
-	timeout time.Duration,
+func ConnectRawSocketPeerContext(
+	ctx context.Context,
 	network,
 	address string,
 	serialization serialize.Serialization,
@@ -83,12 +85,8 @@ func ConnectRawSocketPeerTimeout(
 		return nil, err
 	}
 
-	if timeout == 0 {
-		conn, err = net.Dial(network, address)
-	} else {
-		conn, err = net.DialTimeout(network, address, timeout)
-	}
-
+	var d net.Dialer
+	conn, err = d.DialContext(ctx, network, address)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


transport/rawsocketpeer.go
```diff
@@ -64,7 +64,7 @@ func ConnectRawSocketPeer(network, address string, serialization serialize.Seria
 		return nil, err
 	}
 
-	conn, err := net.Dial(network, address)
+	conn, err := net.DialTimeout(network, address, 5 * time.Second)
```
gammazero: The approach that I would prefer for this is for the caller to pass in a context. That way the caller can set whatever timeout/deadline they want, or call the cancel function stop waiting.  However, that probably requires an API change so need to consider carefully

transport/websocketpeer.go
```diff
@@ -115,7 +116,8 @@ func ConnectWebsocketPeer(routerURL string, serialization serialize.Serializatio
 		dialer.EnableCompression = true
 	}
 
-	conn, _, err := dialer.Dial(routerURL, nil)
+	ctx, _ := context.WithTimeout(context.Background(), 5 * time.Second)
```
gammazero: The approach that I would prefer for this is for the caller to pass in a context. That way the caller can set whatever timeout/deadline they want, or call the cancel function stop waiting.

With this code above, it would still be good to defer-call the returned cancel function so that resources are freed as soon as `ConnectWebSocketPeer()` exits

bkneis: Agreed, this would be nice to pass a context but would break backwards compatibility. How about a ConnectWebsocketPeerContext function that can pass a context, similar to Dial and DialContext? Then I can refactor the ConnectWebsocketPeer to simply call ConnectWebsocketPeerContext with a nil context

transport/rawsocketpeer.go
bkneis: See comment below

gammazero: Yes, like your comment below.  I think that would work well.

bkneis: OK great, will update the PR

transport/websocketpeer.go
gammazero: Yes, that is what I was looking for.  I would like both a `ConnectWebsocketPeerContext` and a `ConnectRawSocketPeerContext`.  If those are provided, then the above code in `ConnectWebsocketPeer` would change to to create a context and call the new `ConnectWebsocketPeerContext` (can defer-call the cancel func).

I will wait for the PR update to introduce those changes before merging anything.

bkneis:
@gammazero The new contextual functions are now added to the PR :)

transport/websocketpeer.go
```diff
@@ -69,14 +70,35 @@ const (
...
 
+// ConnectWebsocketPeer calls ConnectWebsocketPeerContext without a Dial context
+func ConnectWebsocketPeer(
+	routerURL string,
+	serialization serialize.Serialization,
+	tlsConfig *tls.Config,
+	dial DialFunc,
+	logger stdlog.StdLog,
+	wsCfg *WebsocketConfig) (wamp.Peer, error) {
+	return ConnectWebsocketPeerContext(nil, routerURL, serialization, tlsConfig, dial, logger, wsCfg)
```
gammazero: Better to call with first argument as `context.Background()`.  Reason explained below.

transport/websocketpeer.go
```diff
@@ -115,7 +137,12 @@ func ConnectWebsocketPeer(routerURL string, serialization serialize.Serializatio
 		dialer.EnableCompression = true
 	}
 
-	conn, _, err := dialer.Dial(routerURL, nil)
+	if ctx == nil {
```
gammazero: Allowing a nil context instead of using `context.Background()` does not save anything, since `dialer.Dial()` just ends up calling `dialer.DialContext(context.Background(),..)` anyway:
https://golang.org/src/net/dial.go?s=10483:10543#L329

transport/rawsocketpeer.go
```diff
@@ -45,31 +45,55 @@ const (
...
 	if err != nil {
 		return nil, err
 	}
 
-	conn, err := net.Dial(network, address)
+	if timeout == 0 {
+		conn, err = net.Dial(network, address)
+	} else {
+		conn, err = net.DialTimeout(network, address, timeout)
+	}
```
gammazero: Change all this to:
```
var d net.Dialer
conn, err = d.DialContext(ctx,...)
```

transport/rawsocketpeer.go
```diff
@@ -45,31 +45,55 @@ const (
...
 // and address parameters are documented here:
 // https://golang.org/pkg/net/#Dial
 //
 // If recvLimit is > 0, then the client will not receive messages with size
 // larger than the nearest power of 2 greater than or equal to recvLimit.  If
 // recvLimit is <= 0, then the default of 16M is used.
-func ConnectRawSocketPeer(network, address string, serialization serialize.Serialization, logger stdlog.StdLog, recvLimit int) (wamp.Peer, error) {
-	err := checkNetworkType(network)
+func ConnectRawSocketPeerTimeout(
+	timeout time.Duration,
```
gammazero: For symmetry with websocket, please call this `ConnectRawSocketPeerContext` and pass `context.Context` as first arg.

Then, instead of using `net.DialTimeout` do:
```
var d net.Dialer
d.DialContext(ctx,...)
```

transport/rawsocketpeer.go
```diff
@@ -45,31 +45,55 @@ const (
 	magic = 0x7f
 )
 
-// ConnectRawSocketPeer creates a new rawSocketPeer with the specified config,
+// ConnectRawSocketPeer calls ConnectRawSocketPeerTimeout without a Dial timeout
+func ConnectRawSocketPeer(network, address string, serialization serialize.Serialization, logger stdlog.StdLog, recvLimit int) (wamp.Peer, error) {
+	return ConnectRawSocketPeerTimeout(0, network, address, serialization, logger, recvLimit)
```
gammazero: Pass in `context.Background()` as first argument, see comments below.

bkneis:
@gammazero In ConnectWebsocketPeer we are using a net.Conn instead of a Dialer, so DialContext is not available. The only function available other than Dial was DialTimeout, which is why I used it here

gammazero:
@bkneis Please take a look at my updates to your PR to see what I meant about using `DialContext` with rawsocket.

bkneis:
@gammazero Ah OK I see now what you mean, yes that is definitely a better solution so they both use the  Dailer. Is the PR now ready to merge?

gammazero:
@bkneis The PR is now ready.  I needed to add the ability to handle connecting TCP+TLS with Context.

PR #173, opened 2019-04-22
Discard YIELD if it comes from the wrong session

gammazero:
A YIELD message should only come from the session to which the CALL was sent.  If any other session sends a YEILD for that CALL, then the YIELD is discarded and the error is logged.

Added unit test for above.  Made additional fixes to existing unit tests.


gammazero:
Credit to @martin31821 for this fix.

Issue #183, opened 2019-06-25
Client lockup during Register() when session killed


neiledgar:
We are seeing a lockup in the client when the session is closed as the client is registering it's procedures. I could simulate this with an "wamp.session.on_join" procedure which calls "wamp.session.kill" for that session. After a few runs I see client never returns from the Register() method.

I generated a stack dump of the running goroutines and it looks like the client is blocked sending to the actionChan (client/client.go:477) but client run() goroutine has already exited.

With debug enabled the client logs are

$ go run rpc/callee/callee.go -serialize=msgpack
CALLEE> Connected to ws://localhost:8000/ws using msgpack serialization
CALLEE> Client 6882173603170586 received REGISTERED
CALLEE> Registered sum_0 as registration 31
CALLEE> Registered procedure sum_0 with router for client 6882173603170586
CALLEE> Client 6882173603170586 received REGISTERED
CALLEE> Registered sum_1 as registration 32
CALLEE> Registered procedure sum_1 with router for client 6882173603170586
CALLEE> Client 6882173603170586 received REGISTERED
CALLEE> Registered sum_2 as registration 33
CALLEE> Registered procedure sum_2 with router for client 6882173603170586
CALLEE> Client 6882173603170586 received REGISTERED
CALLEE> Registered sum_3 as registration 34
CALLEE> Registered procedure sum_3 with router for client 6882173603170586
CALLEE> Client 6882173603170586 received REGISTERED
CALLEE> Client 6882173603170586 received GOODBYE
CALLEE> Client 6882173603170586 closed

gammazero:
Fixed by #184

PR #184, opened 2019-06-25
Client lockup action chan

neiledgar:
I have made some changed to check the done channel when sending to the action channel. Can you check this is along the lines you expect?

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -276,13 +276,12 @@ func (c *Client) Subscribe(topic string, fn EventHandler, options wamp.Dict) err
 	switch msg := msg.(type) {
 	case *wamp.Subscribed:
 		// Register the event handler for this subscription.
-		sync := make(chan struct{})
-		c.actionChan <- func() {
+		action := func() {
 			c.eventHandlers[msg.Subscription] = fn
 			c.topicSubID[topic] = msg.Subscription
-			close(sync)
 		}
-		<-sync
+		return c.runAction(action)
+
 	case *wamp.Error:
 		return fmt.Errorf("subscribing to topic '%v': %s", topic,
 			wampErrorString(msg))
@@ -296,21 +295,18 @@ func (c *Client) Subscribe(topic string, fn EventHandler, options wamp.Dict) err
 // client does not have an active subscription to the topic, then returns false
 // for second boolean return value.
 func (c *Client) SubscriptionID(topic string) (subID wamp.ID, ok bool) {
-	sync := make(chan struct{})
-	c.actionChan <- func() {
+	action := func() {
 		subID, ok = c.topicSubID[topic]
-		close(sync)
 	}
-	<-sync
+	c.runAction(action)
 	return
 }
 
 // Unsubscribe removes the registered EventHandler from the topic.
 func (c *Client) Unsubscribe(topic string) error {
-	sync := make(chan struct{})
 	var subID wamp.ID
 	var err error
-	c.actionChan <- func() {
+	action := func() {
 		var ok bool
 		subID, ok = c.topicSubID[topic]
 		if !ok {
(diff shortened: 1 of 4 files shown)


neiledgar:
After fixing lockup with action channel, I received a nil pointer exception when a Client.Call() was trying to update the awaitingReply map even though it had been set to nil by the Client.Close() in another go routine

```
2019/06/26 13:33:46 http: panic serving 127.0.0.1:36036: assignment to entry in nil map
goroutine 329 [running]:
net/http.(*conn).serve.func1(0x12d4c120)
        /usr/local/go/src/net/http/server.go:1726 +0x9c
panic(0x641c08, 0x78c6d8)
        /usr/local/go/src/runtime/panic.go:502 +0x204
vsys/mas/vendor/github.com/gammazero/nexus/client.(*Client).expectReply(0x12eec800, 0x3, 0x0)
        /go/src/vsys/mas/vendor/github.com/gammazero/nexus/client/client.go:1029 +0x6c
vsys/mas/vendor/github.com/gammazero/nexus/client.(*Client).CallProgress(0x12eec800, 0x791830, 0x12d7c680, 0x12dfd540, 0x1b, 0x12ee2d40, 0x0, 0x0, 0x0, 0x12ee2ce0, ...)
        /go/src/vsys/mas/vendor/github.com/gammazero/nexus/client/client.go:698 +0x90
vsys/mas/vendor/github.com/gammazero/nexus/client.(*Client).Call(0x12eec800, 0x791830, 0x12d7c680, 0x12dfd540, 0x1b, 0x0, 0x0, 0x0, 0x0, 0x12ee2ce0, ...)
        /go/src/vsys/mas/vendor/github.com/gammazero/nexus/client/client.go:643 +0x7c
```

neiledgar:
I received a second panic when Client.Call() calls Peer.Send() on the session, after the peer has been set to nil in Client.Close() in another goroutine. There is no synchronization between the Close() updating the session and other methods using the session. Also I do not understand why the Session peer is set to nil - the peer is already closed when leaveRealm() method. I have removed the call to set the peer to nil rather than implement synchronization. Is this ok?

```
2019/06/27 08:49:49 http: panic serving 127.0.0.1:49076: runtime error: invalid memory address or nil pointer dereference
goroutine 55824 [running]:
net/http.(*conn).serve.func1(0x1164a0c0)
        /usr/local/go/src/net/http/server.go:1726 +0x9c
panic(0x641bc8, 0xa3bdd8)
        /usr/local/go/src/runtime/panic.go:502 +0x204
vsys/mas/vendor/github.com/gammazero/nexus/client.(*Client).CallProgress(0x117d8680, 0x791840, 0x14e506c0, 0x12e75f20, 0x20, 0x15165f00, 0x0, 0x0, 0x0, 0x15165dc0, ...)
        /go/src/vsys/mas/vendor/github.com/gammazero/nexus/client/client.go:699 +0x128
vsys/mas/vendor/github.com/gammazero/nexus/client.(*Client).Call(0x117d8680, 0x791840, 0x14e506c0, 0x12e75f20, 0x20, 0x0, 0x0, 0x0, 0x0, 0x15165dc0, ...)
        /go/src/vsys/mas/vendor/github.com/gammazero/nexus/client/client.go:643 +0x7c

```

neiledgar:
I have added a unit test to demonstrate the problem. If you are happy with the changes I will squash all the commits into one and reference the issue https://github.com/gammazero/nexus/issues/183

gammazero:
Will be review/testing this weekend.  Thank you very much for you contribution.

gammazero:
The code looks good to me.  I made a few minor changes, mostly cosmetic.  I think I will probably merge this, but first want to consider a somewhat different approach.

neiledgar:
LGTM. Apologies for omitting to run gofmt. 

We use errs.Wrap() locally because it allows the caller to determine the underlying error if required. 
```
	switch errors.Cause(err).(type) {
	case *RPCError:
		// handle specifically
	}
```
However, I agree it is not required here, given NetConnect() is not returning any types error classes.

Issue #189, opened 2019-07-17
`client.Close` sometimes hangs on network error


tomkcook:
I have a situation where the network upstream of the device running nexus code is unreliable.  When the network connection breaks, it does so without anything telling my nexus-based client that it's broken (ie no TCP FIN or RST packets are sent).  In this case, about one time in two, I see my client hanging in `client.Close()` with this included in the stack (note that line numbers here are for commit 4debb4a but I believe the same defect is present in the current master):

```
Jul 15 12:03:09 C05AGB0000A000000297 daemon.info docker:ies-system-armhf2.2.0-87/2ec45310b45a[2257]: goroutine 1 [semacquire, 31 minutes]:
Jul 15 12:03:09 C05AGB0000A000000297 daemon.info docker:ies-system-armhf2.2.0-87/2ec45310b45a[2257]: sync.runtime_Semacquire(0x11a9c560)
Jul 15 12:03:09 C05AGB0000A000000297 daemon.info docker:ies-system-armhf2.2.0-87/2ec45310b45a[2257]: 	/usr/local/go/src/runtime/sema.go:56 +0x2c
Jul 15 12:03:09 C05AGB0000A000000297 daemon.info docker:ies-system-armhf2.2.0-87/2ec45310b45a[2257]: sync.(*WaitGroup).Wait(0x11a9c554)
Jul 15 12:03:09 C05AGB0000A000000297 daemon.info docker:ies-system-armhf2.2.0-87/2ec45310b45a[2257]: 	/usr/local/go/src/sync/waitgroup.go:129 +0x84
Jul 15 12:03:09 C05AGB0000A000000297 daemon.info docker:ies-system-armhf2.2.0-87/2ec45310b45a[2257]: vsys/mas/vendor/github.com/gammazero/nexus/client.(*Client).Close(0x11a9c500, 0x11fbbc10, 0x1)
Jul 15 12:03:09 C05AGB0000A000000297 daemon.info docker:ies-system-armhf2.2.0-87/2ec45310b45a[2257]: 	/go/src/vsys/mas/vendor/github.com/gammazero/nexus/client/client.go:760 +0x60
```

As far as I can tell, this is what's happening:

* On calls to `runHandleInvocation`, `c.activeInvHandlers.Add(1)` is called and then a goroutine is kicked off to handle the invocation.  The first thing this does is to `defer` this function:
```
		defer func() {
			c.progGateLock.Lock()
			delete(c.progGate, ctx)
[...]

PR #190, opened 2019-07-17
Turn `runAction` into a promise-ish pattern.

tomkcook:
This allows functions using `runAction` to specify what should happen
after the requested function has succeeded or failed.  This is then used
to clean up a leak of active invocation handlers - see issue #189 for a
complete description.

Also slightly rejigged how error codes are returned to the users of
`runAction` - this allows actions started with `runAction` to
complete asynchronously if that is what is desired.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -276,10 +276,11 @@ func (c *Client) Subscribe(topic string, fn EventHandler, options wamp.Dict) err
 	switch msg := msg.(type) {
 	case *wamp.Subscribed:
 		// Register the event handler for this subscription.
-		return c.runAction(func() {
+		return c.runAction(func() error {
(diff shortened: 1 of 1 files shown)


tomkcook:
I've checked in the field that this resolves my problem.

gammazero:
Your approach, with functionality to supply callbacks that are executed on success or failure, is more than is needed currently IMO. While it offers much flexibility, I would prefer to keep the code simpler and avoid any GC pressure unless the functionality is really needed.

I really appreciate you finding, analyzing, and fixing this issue.  I also think your approach is a useful general concurrency pattern.  However, it is not quite the right time to employ it here yet, so I closing this PR.  Again, thank you.

PR #191, opened 2019-07-23
Fix issue #189

gammazero:
Always signal when an invocation handler has finished.  Includes unit test of router that disconnects callee during invocation handling.


gammazero:
Credit to @tomkcook for identifying the problem (#189) and for supplying the initial fix.

gammazero:
Relaced by #193

PR #199, opened 2019-09-09
Feature/add debug print

bkneis:
DialContext is capable of returning a non nil response after returning an error, this response is critical for debugging connection errors.

A complete example of this is when an ErrBadHandshake is returned. The error string simply shows "websocket: bad handshake", without access to the response struct this is very difficult to debug. This PR returns the response so it can be printed by the caller

### DIFF
--- a/client/network.go
+++ b/client/network.go
@@ -56,7 +56,7 @@ func ConnectNetContext(ctx context.Context, routerURL string, cfg Config) (*Clie
 		routerURL = u.String()
 		fallthrough
 	case "ws", "wss":
-		p, err = transport.ConnectWebsocketPeerContext(ctx, routerURL, cfg.Serialization,
+		p, _, err = transport.ConnectWebsocketPeerContext(ctx, routerURL, cfg.Serialization,
 			cfg.TlsCfg, cfg.Dial, cfg.Logger, &cfg.WsCfg)
 	case "tcp":
 		p, err = transport.ConnectRawSocketPeerContext(ctx, u.Scheme, u.Host,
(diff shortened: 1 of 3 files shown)


client/network.go
```diff
@@ -56,7 +56,7 @@ func ConnectNetContext(ctx context.Context, routerURL string, cfg Config) (*Clie
 		routerURL = u.String()
 		fallthrough
 	case "ws", "wss":
-		p, err = transport.ConnectWebsocketPeerContext(ctx, routerURL, cfg.Serialization,
+		p, _, err = transport.ConnectWebsocketPeerContext(ctx, routerURL, cfg.Serialization,
```
gammazero: Seems like something should be done with the response here:
1. Handle redirect
2. Include HTTP status and response body, or whole response, in error

Consider a new type, `transport.WebSockerError` that can be returned as an `error`:
```
type WebsockerError struct {
    Err error
    Response *http.Response
}
func (e *WebsockerError) Error() string {
    if e.Response == nil {
        return e.Err.Error()
    }
    return fmt.Sprintf("%s: %s", e.Err, e.Response.Status)
}
```

This allows the caller to cast the error to retrieve the http response:
```
if wsErr, ok := err.(*transport.WebsocketError); ok {
    rsp := wsErr.Response
    // do something with rsp
}
```
This error should come from the transport code.

transport/websocketpeer.go
```diff
@@ -128,19 +129,19 @@ func ConnectWebsocketPeerContext(ctx context.Context, routerURL string, serializ
...
 		}
 		dialer.Jar = wsCfg.Jar
 		dialer.EnableCompression = true
 	}
 
-	conn, _, err = dialer.DialContext(ctx, routerURL, nil)
+	conn, res, err = dialer.DialContext(ctx, routerURL, nil)
 	if err != nil {
-		return nil, err
+		return nil, res, err
```
gammazero: Consider returning a `*WebsocketError` here, as discussed above.
```
return nil, &WebsocketError{
    Err: err,
    Response: res,
}
```

gammazero:
Closing in favor of #200 implementation.  Thank you for contributing this.  PR #200 comment give s credit to you.

PR #209, opened 2019-11-26
apply filter only if argument is not nil

sonork:
Bugfix in implementation for wamp.session.list
Optional filter argument is present but nil on requests received from autobahn python client.


gammazero:
I think this means a more general fix is probably better.  Instead of fixing the specific request handler, I think that trying to convert `nil` to `wamp.List` should be OK, but return a `nil` `wamp.List`

gammazero:
Closing: replaced by #213

PR #217, opened 2020-01-22
[NFR] Allow custom connections to be specified in transport.NewWebsocketPeer()

lajosbencz:
It would be nice to be able to run your client library embedded in a WASM binary.
The problem is that TCP connections are not supported, so the https://github.com/gorilla/websocket implementation is a no-go...

This PR would allow for custom connections to be specified in ```transport.NewWebsocketPeer```, like a wrapper around ```window.WebSocket``` using the js runtime.

### DIFF
--- a/router/websocketserver.go
+++ b/router/websocketserver.go
@@ -344,7 +344,7 @@ func (s *WebsocketServer) addProtocol(proto string, payloadType int, serializer 
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


lajosbencz:
Some of the version specific tests have failed, but as far as i understand this change shouldn't brake anything. What have I done wrong? :D

PR #234, opened 2020-11-29
Reduce memory allocation

gammazero:


client/client.go
```diff
@@ -69,6 +69,8 @@ type InvokeResult struct {
 // the invocation was canceled.
 var InvocationCanceled = InvokeResult{Err: wamp.ErrCanceled}
 
+var emptyDict = wamp.Dict{}
```
dcarbone, line 72: Unless I'm missing something, a side effect of this is that if an implementor has connected a `localPeer` to any realm modifies an `options`, `details`, etc. field in a received message that has been set to `emptyDict`, the `emptyDict` var itself would be updated.

This could quickly lead to concurrent write panics, as well.  Later today I can build a test case for this to be sure, but I wanted to point it out just in case it'd been overlooked :)

gammazero: Yes, you are correct.  There was an assumption that local clients would not alter message contents, and being local clients that are trusted, they can be trusted to behave.  Maybe this is not a safe assumption.

If we want to allow local clients to alter message contents, this also means there will need to be some changes to things like publishing the same event to multiple clients, or calling RPC that goes to multiple clients.

Do you prefer to allow local clients to alter message contents, at the expense of having to make copies of data, or do you thing is it reasonable to have a rule that states that a local client may not alter message content as that may corrupt the message received by other clients?

PR #236, opened 2020-11-29
emptyDict leak test

dcarbone:
Test to prove the dangerous nature of the new `emptyDict` implementation.


dcarbone:
sorry for the commit mess.  its early here yet :)

PR #238, opened 2020-11-30
Add HeapAlloc to logMemStats

benk79:


router/router.go
```diff
@@ -104,7 +104,7 @@ func (r *router) logMemStats(interval time.Duration) {
 	for {
 		time.Sleep(interval)
 		runtime.ReadMemStats(&m)
-		r.log.Printf("MemStats: Alloc=%d Mallocs=%d Frees=%d NumGC=%d", m.Alloc, m.Mallocs, m.Frees, m.NumGC)
+		r.log.Printf("MemStats: Alloc=%d Mallocs=%d Frees=%d NumGC=%d HeapAlloc=%d", m.Alloc, m.Mallocs, m.Frees, m.NumGC, m.HeapAlloc)
```
gammazero, line 107: The documentation says that `Alloc` is the same as `HeapAlloc`: https://golang.org/pkg/runtime/#MemStats
Is this incorrect?
Also, I never saw these report different values (at least not with go1.15).

gammazero (commented):
I thought `Alloc` and `HeapAlloc` are the same, according to documentation: https://golang.org/pkg/runtime/#MemStats

gammazero:
I think this PR can be closed with the explanation here https://github.com/gammazero/nexus/pull/238#discussion_r532741932

Issue #245, opened 2021-04-24
client connection gets dropped due to it being unable to reply to a socket ping


DewitteRuben:
Setup:
- Raspberry Pi running the WAMP client and Docker, connected to a Crossbar instance
- RPC calls are used to start/stop docker builds/pulls/runs

It seems that, from what I've seen, that the client is blocked and with that unable to reply to a ping from the socket in question (a crossbar instance in this case). This causes the client to receive a Done signal forcing me to re-establish the socket connection.

It seems to only occur when the client busy in some kind of way

Some ideas I had that might cause this issue:
- Client is busy handling doing something (e.g. handling an RPC call), and is unable to respond to the ping
  - The issue only seems to occur whenever an additional RPC is being called, so maybe the client is being blocked in some way or another?
- Device runs out of resources and is unable to respond to the ping
  - Kind of doubt this reason since upon viewing CPU usage of device, it seems to have CPU / RAM to spare

Things we have tried, besides going over the code multiple times:
- Setting niceness to the executable running the WAMP client
- Shaping the network traffic to prioritize that of the client
- Making sure nothing is blocking an RPC handler / the client in any way

Any kind of help or hints would be appreciated..

gammazero:
If you have control of the server, disabling websocket keep-alives may be a temporary solution.  Seems similar to this: https://github.com/gorilla/websocket/issues/97

So, I may need to replace the ping handler as described [here](https://github.com/gorilla/websocket/issues/97#issuecomment-194967563).

DewitteRuben:
@gammazero Thanks for the reply!

Sadly, the temporary solution of disabling the keep-alive is not an option for us. We currently use the keep-alives to toggle a 'device' it's connection status in our database. So we need our crossbar instance to be aware if one of our clients has an active connection or not.

The issue you're referring would explain the behavior we are experiencing, might take a look into that to see if that's actually the cause.

gammazero:
If you can try PR #247 and see if that fixes the situation for you, then I can merge that if that is the solution.

DewitteRuben:
@gammazero Thanks a lot for the effort, and reaction time.

Unfortunately it seems that this did not resolve the issue. With further investigation it seems like this issue does not occur when there the internet connection is established with ethernet, so it might just be related to some OS configuration, the Raspberry Pi's WiFi hardware, or maybe a power issue under heavy CPU load. I also can't reproduce the issue when trying the executable on my mac.

[...]

PR #247, opened 2021-04-28
Provide ping handler that responds when client blocked

gammazero:
The websocket only handles responding to pings when reading messages.  If the client is blocked handling incoming messages and does not read additional messages for some time, websocket keep-alive pings are not responded to.  This may lead to the server dropping the websocket connection thinking that the client is dead.  This PR provides a ping handler that answers pings even when the client is not reading messages.

Addresses issue #245


om26er:
I think this issue still exists. I have a WAMP client calling into my device, the invocation result in some cases can take quite a few seconds. The websocket connection is hence dropped

Issue #267, opened 2022-04-12
Override caller details for relay mode


om26er:
I have implemented router to router links based on top of Nexus https://github.com/om26er/nexlinks. It extends the Nexus router to act as a relay i.e. forward all WAMP procedures registered on one router to another router.

This is especially suitable for edge-cloud scenarios. One issue I see is that when a message is relayed between two routers, the caller details are wrong, this is because the relay itself is ultimately the caller.

We should support the nexus dealer to override that behavior

om26er:
I have put out a proposal for this issue. We would also need to do a similar treatment in the broker code for publish events.

PR #268, opened 2022-04-12
allow to override caller/publisher details when acting as relay

om26er:
A proposal for https://github.com/gammazero/nexus/issues/267, your feedback is welcome @gammazero

### DIFF
--- a/router/broker.go
+++ b/router/broker.go
@@ -497,7 +497,7 @@ func (b *broker) syncPubEvent(pub *wamp.Session, msg *wamp.Publish, pubID wamp.I
 			event.Details[detailTopic] = msg.Topic
 		}
 		if disclose && subscriber.HasFeature(wamp.RoleSubscriber, wamp.FeaturePubIdent) {
-			disclosePublisher(pub, event.Details)
+			disclosePublisher(pub, msg, event.Details)
 		}
 
 		if subscriber.Peer.IsLocal() {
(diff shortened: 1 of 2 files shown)


om26er:
Here is a snippet of how the in-router client could call this
```
func RegisterRemote(uri string, regID int, regMap *sync.Map, logger *logrus.Logger) {
	invocationHandler := func(ctx context.Context, inv *wamp.Invocation) client.InvokeResult {

		options := wamp.Dict{}
		if id, ok := wamp.AsID(inv.Details["caller"]); ok {
			options["disclose_me"] = true

			// All x_nexus_* parameters are specific to nexus, we use
			// these to tell the dealer that the CALL is actually a
			// relay call, hence the actual CALLER should be forwarded
			// to the call.
			// FIXME: needs an out-of-tree patch to nexus
			options["x_nexus_session_id"] = id
			options["x_nexus_authrole"] = inv.Details["caller_authrole"]
			options["x_nexus_authid"] = inv.Details["caller_authid"]
		} else {
			options["disclose_me"] = false
		}

		response, err := localSession.Call(ctx, uri, options, inv.Arguments, inv.ArgumentsKw, nil)
		if err == nil {
			return client.InvokeResult{Args: response.Arguments, Kwargs: response.ArgumentsKw}
		} else {
			realError := err.(client.RPCError).Err
			return client.InvokeResult{Args: realError.Arguments, Kwargs: realError.ArgumentsKw,
				Err: realError.Error}
		}
	}

	options := wamp.Dict{}
	options["disclose_caller"] = true
	err := remoteSession.Register(uri, invocationHandler, options)
	if err == nil {
		regMap.Store(regID, uri)
		logger.Printf("Registered procedure on remote %s", uri)
	} else {
		logger.Errorln(err)
	}
}
```

om26er:
Updated this PR to use `forward_for` dictionary, which is based on same approach that Crossbar uses. For example https://github.com/crossbario/crossbar/blob/cce29666fd0c173a41ea6e33c7e19b8a0427458b/crossbar/router/dealer.py#L953

om26er:
Closing the PR as it's not part of the spec. To actually test this feature I have forked Nexus.

Will come back with a similar PR once RLinks becomes part of the WAMP spec.

Issue #260, opened 2022-02-15
Support multiple serializers


om26er:
Currently a client has to provide the serializer that it wants to use when connecting to WAMP router. Nexus should support providing multiple WAMP serializers, just like autobahn and other WAMP clients.

om26er:
To an extent, that supported now

PR #270, opened 2022-06-08
Client: offer all supported serializers

om26er:
Without changing the user API, takes care of https://github.com/gammazero/nexus/issues/260

If no serializer is provided in the config: offer all supported
If specific serializer is request, keep the existing behavior

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -691,7 +691,7 @@ func (ks *serverKeyStore) AuthRole(authid string) (string, error) {
 
 func TestConnectContext(t *testing.T) {
 	const (
-		expect     = "dial tcp: operation was canceled"
+		expect     = "dial tcp: lookup localhost: operation was canceled"
 		unixExpect = "dial unix /tmp/wamp.sock: operation was canceled"
 	)
 
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -691,7 +691,7 @@ func (ks *serverKeyStore) AuthRole(authid string) (string, error) {
 
 func TestConnectContext(t *testing.T) {
 	const (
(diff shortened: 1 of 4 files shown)


om26er:
ping @gammazero, can you please take a look at this one ?

Also I'd like to help with the maintenance of Nexus. I have a bunch of features planned that I'd like to implement

gammazero (approved):
This looks good, and I intend to merge it after a few more tests.

As far as maintaining this code, I am considering a fairly significant overhaul.  Many of the APIs and configurations can be simplified, as well as changes to the internal queuing.

client/client_test.go
```diff
@@ -691,7 +691,7 @@ func (ks *serverKeyStore) AuthRole(authid string) (string, error) {
 
 func TestConnectContext(t *testing.T) {
 	const (
-		expect     = "dial tcp: lookup localhost: operation was canceled"
+		expect     = "dial tcp: operation was canceled"
```
gammazero, line 694: I think the better way to fix this is to either change CI to use a later version of go, or have the test look for "operation was canceled"

gammazero (approved):

om26er:
> As far as maintaining this code, I am considering a fairly significant overhaul. Many of the APIs and configurations can be simplified, as well as changes to the internal queuing.

Right. The authentication API in both the client and server API can be simplified. I have already made some progress on that front, will share the changes I made.

om26er:
Regarding internal queuing. I believe that are some bottlenecks around concurrency, in my benchmarks, the RPC/second count doesn't gain significantly as the number of goroutines increases.

On a single core I could do 40k+ RPC/second (empty payload) but as the concurrency is increased, we can clearly see diminishing returns

PR #274, opened 2022-10-03
Payload PassThru Mode

KSDaemon:
This PR introduces WAMP AP Feature `Payload PassThru Mode` (PPT for short).

This feature is implemented in all 6 roles. Also covered with tests.


KSDaemon:
Fixed problematic tests (actually bugs, not the tests itself :) )

KSDaemon:
@gammazero Hi! Just a gentle reminder about PR :)

oberstet:
From a quick look (obviously, @gammazero should say ..), this looks quite "complete". My only "worry": PPT mode is required as basis for both MQTT and E2EE support - and only [PPT-MQTT](https://github.com/crossbario/crossbar-examples/tree/master/mqtt/basic#payload-mapping) or PPT-E2EE would allow creating interoperability tests with AutobahnPython / Crossbar.io. Or @KSDaemon do you have any ideas rgd (interop) testing (vs the spec / other impls) of PPT alone?

KSDaemon:
Regarding interop tests — I think we can create another repo in our gh org just for that. As crossbar is already dockerized, and it won't be hard to dockerize nexus, and no problems with clients. We can start creating such a complex test with every client, serializer, mode, features.

client/client.go
```diff
@@ -579,6 +799,48 @@ func (c *Client) Call(ctx context.Context, procedure string, options wamp.Dict,
...
+				abortMsg := wamp.Abort{
+					Reason: wamp.ErrProtocolViolation,
+					Details: wamp.Dict{
+						wamp.OptError:   ErrPPTNotSupportedByRouter.Error(),
+						wamp.OptMessage: ErrPPTNotSupportedByPeer.Error(),
+					},
+				}
+				c.sess.Peer.Send(&abortMsg)
+				c.sess.Peer.Close()
+				return nil, fmt.Errorf("cannot process unexpected passthrough result: %v", ErrPPTNotSupportedByRouter)
```
gammazero, line 818: It probably makes sense to wrap errors, so that the type of error can be tested at a higher level.
```suggestion
				return nil, fmt.Errorf("cannot process unexpected passthrough result: %w", ErrPPTNotSupportedByRouter)
```

This is something that can likely be done throughout the code, and changes to code outside of this PR can be done separately.

client/client.go
```diff
@@ -1108,6 +1370,41 @@ func (c *Client) runHandleEvent(msg *wamp.Event) {
...
+	// Check and handle Payload PassThru Mode
+	// @see https://wamp-proto.org/wamp_latest_ietf.html#name-payload-passthru-mode
+	// Since it is async publication, and no link to
+	// original publication - as it was already published
+	// we can not reply with error, only log it.
+	// Although the router should handle it
+	if pptScheme, _ := msg.Details[wamp.OptPPTScheme].(string); pptScheme != "" {
+
+		if !isPPTSchemeValid(pptScheme) {
+			c.log.Print(fmt.Sprintf("cannot process even with invalid ppt schema %q: %v", pptScheme, ErrPPTSchemeInvalid))
```
gammazero, line 1383:
```suggestion
			c.log.Printf(cannot process even with invalid ppt schema %q: %v", pptScheme, ErrPPTSchemeInvalid)
```

client/client.go
```diff
@@ -1108,6 +1370,41 @@ func (c *Client) runHandleEvent(msg *wamp.Event) {
...
+		// Now need to check ppt_serializer (in pair with ppt_scheme)
+		// and serialize payload with appreciate serializer
+		if pptScheme == "wamp" {
+			args, kwargs, err = unpackE2EEPayload(msg.Details, msg.Arguments)
+		} else {
+			args, kwargs, err = unpackPPTPayload(msg.Details, msg.Arguments)
+		}
+
+		if err != nil {
+			c.log.Print(fmt.Sprintf("cannot unpack event: %v", err))
```
gammazero, line 1400:
```suggestion
			c.log.Printf("cannot unpack event: %v", err)
```

client/client.go
```diff
@@ -1140,6 +1437,50 @@ func (c *Client) runHandleInvocation(msg *wamp.Invocation) {
...
+
+		if !isPPTSchemeValid(pptScheme) {
+			c.sess.Send(&wamp.Error{
+				Type:      wamp.INVOCATION,
+				Request:   reqID,
+				Details:   wamp.Dict{},
+				Error:     wamp.ErrInvalidArgument,
+				Arguments: wamp.List{ErrPPTSchemeInvalid.Error()},
+			})
+			c.log.Print(fmt.Sprintf("cannot process invocation with invalid ppt schema %q: %v", pptScheme, ErrPPTSchemeInvalid))
```
gammazero, line 1452:
```suggestion
			c.log.Printf("cannot process invocation with invalid ppt schema %q: %v", pptScheme, ErrPPTSchemeInvalid)
```

client/client.go
```diff
@@ -1140,6 +1437,50 @@ func (c *Client) runHandleInvocation(msg *wamp.Invocation) {
...
+
+		if err != nil {
+			c.sess.Send(&wamp.Error{
+				Type:      wamp.INVOCATION,
+				Request:   reqID,
+				Details:   wamp.Dict{},
+				Error:     wamp.ErrInvalidArgument,
+				Arguments: wamp.List{err.Error()},
+			})
+			c.log.Print(fmt.Sprintf("cannot unpack invocation message: %v", err))
```
gammazero, line 1476:
```suggestion
			c.log.Printf("cannot unpack invocation message: %v", err)
```

router/broker.go
```diff
@@ -483,12 +506,17 @@ func (b *broker) syncPubEvent(pub *wamp.Session, msg *wamp.Publish, pubID wamp.I
 
 		// TODO: Handle publication trust levels
 
+		details := wamp.Dict{}
+		if eventDetails != nil {
+			details = eventDetails
+		}
```
gammazero, line 512: Nice

gammazero (approved):
This is excellent. Thank you very much. There are a few minor code comments, but nothing blocking.  I will give you permission to merge and you can do that after any final changes you want to make. 🎉

client/client.go
KSDaemon: yeah! I want to setup golint-ci and fix most warnings.

router/broker.go
KSDaemon: :) That was also needed :)

KSDaemon:
So all notes are fixed. Tests are passing. Merging!

PR #276, opened 2022-10-12
Ci Workflow updates

KSDaemon:
What's included: 

- Templates for issues and PRs
- Golang ci linter
- A few more checks
- Moved from travis → github actions
- Fixed a bunch of go linter warnings (but still more exists)


KSDaemon:
No, there is a weird flow of git linter of PR title/Description: it doesn't see [sometimes] actual description and check the one at the creation time. That's seems to be not a problem if we'll get templates in place

PR #278, opened 2022-10-12
Ci Workflow updates

### LINKED ISSUE #279: Convert from travisCI to github workflows
Use github workflows to automate testing.

KSDaemon:
What's included: 

- Templates for issues and PRs
- Golang ci linter
- A few more checks
- Moved from travis → github actions
- Fixed a bunch of go linter warnings (but still more exists)

Fixes #279 


client/client_test.go
```diff
@@ -314,7 +314,7 @@ func TestSubscribe(t *testing.T) {
 	// Make sure the event was not received.
 	select {
 	case <-time.After(time.Millisecond):
-	case err = <-errChan:
+	case <-errChan:
```
gammazero, line 317: 👍 Ther is much more of this through out the code base. It is probably worth a follow-up PR just to get rid of all the lint warnings.

client/client.go
```diff
@@ -711,7 +714,13 @@ type ProgressHandler func(*wamp.Result)
...
 // sufficient for the caller to receive all progressive results as well as the
 // final result.
-func (c *Client) Call(ctx context.Context, procedure string, options wamp.Dict, args wamp.List, kwargs wamp.Dict, progcb ProgressHandler) (*wamp.Result, error) {
+func (c *Client) Call(
+	ctx context.Context,
+	procedure string,
+	options wamp.Dict,
+	args wamp.List,
+	kwargs wamp.Dict,
+	progcb ProgressHandler) (*wamp.Result, error) {
```
gammazero, line 723: I prefer not to put each argument on a separate line. It leads to different contributors making different choices about when to split a function args to separate lines, and can make the code look inconsistent. I also prefer to keep more aligned with the golang stdlib code.

I think the long lines also act as a flag telling us that function needs to be examined, and a better interface is probably needed.

I would prefer to keep the func signature on one line, unless you strongly feel this greatly improves readability. Either way, in a future version of nexus we can probably com up with a better interface for calling functions like this.

gammazero (approved):
A couple of minor comments, but this looks good. Thank you.

Please **squash-merge** when you merge so that each merge to main (v3) has a single commit with all the comments in it.

I had actually started on doing this earlier, but you are first with a PR. I think we should open an issue about what we are going to work and assign it to ourselves. That way we know what eachother is working on.

gammazero (approved):
LGTM

KSDaemon:
So we still have lint warnings:
* 23 testpackage
* 13 gochecknoglobals
* 13 errorlint
* 50 errcheck. Most lacking of Err checking. Sometimes it is not needed (like for connection closing)

gammazero (approved):
I am happy to fix remaining lint warnings in a follow-up PR.  Many of the error checks come from not checking the result of calls that only return a nil error.

Feel free to merge this if you are ready.

KSDaemon:
That will be great!

PR #283, opened 2022-11-02
Implement HasRealm() to know if a realm is already started

om26er:
In some cases it is useful to check if a realm is running on a router. This PR helps with that.

- [ ] Bug fix (non-breaking change which fixes an issue)


KSDaemon:
@om26er Great! A pair of false/positive tests will be a great addition to this PR! ;)

PR #296, opened 2023-04-27
Dont crash if get_events called without args

om26er:
Fixes a crash when `wamp.subscription.get_events` is called without any arguments

(You can also link to an open issue here)

### What is the new behavior?
(if this is a feature change)

### Checklist:

- [ ] My code follows the code style of this project.


router/broker_test.go
```diff
@@ -1311,4 +1311,23 @@ func TestEventHistory(t *testing.T) {
 	if ev1.Arguments[0].(int) < ev2.Arguments[0].(int) {
 		t.Fatalf("MetaRPC subEventHistory for topic %s should return records in reverse order", topic)
 	}
+
```
KSDaemon, line 1314: You forgot to add the `reqId++` here. So the invocation below will be treated as a new request.

KSDaemon:
Great! Thnx for fixing!

PR #297, opened 2023-07-07
WIP: Progressive call invocations sync with the WAMP Spec

KSDaemon:

After initial progressive call invocation implementation the WAMP Spec was updated: we renamed the feature and described some important gaps in the flow. 
So Nexus implementation needs to be updated.

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Checklist:


router/dealer.go
```diff
@@ -646,6 +646,18 @@ func (d *dealer) syncCall(caller *wamp.Session, msg *wamp.Call) {
...
+		// The Caller did not announce the progressive call invocations feature during the HELLO handshake.
+		abortMsg := wamp.Abort{Reason: wamp.ErrProtocolViolation}
+		abortMsg.Details = wamp.Dict{}
+		abortMsg.Details[wamp.OptMessage] = "Peer is trying to use Progressive Call Invocations while it was not " +
+			"announced during HELLO handshake"
+		caller.Peer.Send(&abortMsg)
+		//caller.Peer.Close()
+
+		return
+	}
```
KSDaemon, line 659: @gammazero hi! I need your help a bit! 

During fix/testing I faced a problem: there are some cases when I need to abort peer session from the `sync*` methods. Like here: client is making an RPC call (`func (d *dealer) call(caller *wamp.Session, msg *wamp.Call) `)  which in turn pushes a func into the dealer's action channel:
```
	d.actionChan <- func() {
		d.syncCall(caller, msg)
	}
```

And within `syncCall` in some cases, we fall into protocol violation and the session needs to be aborted. It looks like I'm doing it wrong (see the code above).... so I'm receiving: `panic: send on closed channel`. 

Can't figure out how to make that properly.... Maybe you can suggest how?
There is no problem with doing it from the client side outside the dealer, but from here - I'm stuck...

router/dealer.go
```diff
@@ -646,6 +646,18 @@ func (d *dealer) syncCall(caller *wamp.Session, msg *wamp.Call) {
...
+		// The Caller did not announce the progressive call invocations feature during the HELLO handshake.
+		abortMsg := wamp.Abort{Reason: wamp.ErrProtocolViolation}
+		abortMsg.Details = wamp.Dict{}
+		abortMsg.Details[wamp.OptMessage] = "Peer is trying to use Progressive Call Invocations while it was not " +
+			"announced during HELLO handshake"
+		caller.Peer.Send(&abortMsg)
+		caller.Peer.Close()
+
+		return
+	}
```
KSDaemon, line 659: @gammazero hi! I need your help a bit! 

During fix/testing I faced a problem: there are some cases when I need to abort peer session from the `sync*` methods. Like here: client is making an RPC call (`func (d *dealer) call(caller *wamp.Session, msg *wamp.Call) `)  which in turn pushes a func into the dealer's action channel:
```
	d.actionChan <- func() {
		d.syncCall(caller, msg)
	}
```

And within `syncCall` in some cases, we fall into protocol violation and the session needs to be aborted. It looks like I'm doing it wrong (see the code above).... so I'm receiving: `panic: send on closed channel`. 

Can't figure out how to make that properly.... Maybe you can suggest how?
There is no problem with doing it from the client side outside the dealer, but from here - I'm stuck...

KSDaemon: This is done in the same fashion as in router `attachClient`:
```
func (r *router) AttachClient(client wamp.Peer, transportDetails wamp.Dict) error {
	sendAbort := func(reason wamp.URI, abortErr error) {
		abortMsg := wamp.Abort{Reason: reason}
		abortMsg.Details = wamp.Dict{}
		if abortErr != nil {
			abortMsg.Details[wamp.OptMessage] = abortErr.Error()
			r.log.Println("Aborting client connection:", abortErr)
		}
		client.Send(&abortMsg) // Blocking OK; this is session goroutine.
		client.Close()
	}
```

IDK, maybe that problem is specific only for a LocalPeers (used in tests) cause I faced a few times when some behavior was appearing only in localPeer case but not in SocketPeers.

KSDaemon: So the problem is that `client.Send(&abortMsg)` only puts the message into the queue which is processed in a separate goroutine. But the `client.Close()` is called immediately. I'm wondering is that working in the router context?... Am I missing something?...

KSDaemon:
@gammazero gentle pinging...

KSDaemon:
Okay. As all the current tests are passed I'm merging. Fix can be made later.

gammazero:
I will look into this this week. Thanks for handling this so far.

KSDaemon:
Great! Thank you!

Issue #298, opened 2023-08-05
[Question] Nexusd container image


ivanov17:
Hello! Are you planning to provide official container image for nexusd router?

Thank you.

ivanov17:
Thank you for the quick response. I will think about it. 

I currently need a WAMP router to run a Buildbot multi-master setup. I tried Crossbar.io as recommended by the Buildbot documentation, but it seems like overkill for my purposes. I would like to use a simpler/lightweight solution.

So, if no one will provide nexusd container image, maybe I'll give it a try, but not promises. Of course, I would prefer a ready-made solution, but I don't want to close the issue yet.

ivanov17:
I have to say that nexusd is great. 

For over a week I've been trying to get the official Crossbar container to work with the Buildbot master instances. I tried to run it with different config options, but to no avail. My questions to its developers remained unanswered. In addition, it refused to work due to lack of memory and requires more than 2 Gb for the image.

Today I built nexusd container image. First, it requires only 16 Mb. Second, I only had to specify two parameters for it to work with my Buildbot instances. It's amazing.

I would be happy to contribute to the project and provide a Dockerfile.

KSDaemon:
That's awesome!

PR #299, opened 2023-08-18
Official nexusd container image

ivanov17:

Provides a Dockerfile for `nexusd` container, Github Actions instructions for building the container image, and adds help on using the container to the README file.

Fixes <https://github.com/gammazero/nexus/issues/298>.

There is no official `nexusd` container image.

### What is the new behavior?

The official `nexusd` container image would be provided on the GitHub container registry. The latest version will be rebuilt for every merge in the **v3** branch, numbered versions should be built when tag is added. Binary is built using the latest Go container from the Docker team. The distribution container is based on a tiny busybox image, it is small and easy to use. Instructions for using the container should be added to the README file.

Container has been tested with a Buildbot multimaster installation and works perfectly. It can be run as a privileged or unprivileged process, and it can work with a mounted configuration file.

I'm not sure if any tests need to be added here, but if they are, please let me know.

- [ ] Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [ ] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
  - [x] I have updated the documentation accordingly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM docker.io/library/golang:latest AS build-image
+
+ARG BUILD_VERSION="latest"
+
+RUN go install github.com/gammazero/nexus/v3/nexusd@${BUILD_VERSION}
+
+FROM docker.io/library/busybox:latest AS release-image
+
+ARG BUILD_VERSION="latest"
+
+COPY --from=build-image /go/bin/nexusd /usr/bin/nexusd
+
+EXPOSE 8080
+
+ENTRYPOINT ["/usr/bin/nexusd"]
+CMD ["--ws", "0.0.0.0:8080", "--realm", "realm1"]
+
+ENV TZ=Etc/UTC
+
+LABEL name="nexusd"
+LABEL version="${BUILD_VERSION}"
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM docker.io/library/golang:latest AS build-image
+
+ARG BUILD_VERSION="latest"
+
+RUN go install github.com/gammazero/nexus/v3/nexusd@${BUILD_VERSION}
+
+FROM docker.io/library/busybox:latest AS release-image
+
+ARG BUILD_VERSION="latest"
```
KSDaemon, line 9: Do we need the second arg definition? As they are the same. Whenever we need to distinguish them - we can always introduce new arg. What do you think?

KSDaemon (approved):
👍🏻  This is awesome! Thank you! 
Just a small comment.

Dockerfile
KSDaemon: O thnx for the detailed description, my knowledge in docker/containers seems to be not so deep!

KSDaemon:
👍🏻 So no reason to prevent merging!

PR #300, opened 2023-09-06
Minor Fixes and Improved Test Code

gammazero:
## Bug fixes
- Calling `router.Close` stops the `logMemStats` goroutine if it is running.
- Realm marks `sessionHandler` as done (`WaitGroup.Done()`) at the end of the goroutine
- Fix rawsocket not closing correctly after sending last message
- Fix missed error check when decoding binary json data

_The defects fixed above should not lead to any noticeable change in the behavior of the router or client._

## Improve test code

**Test Improvements**
- Improve goroutine leak detection
- Consolidate common cleanup code
- Clarify tests with `require` test package

**Router Improvements**
- Cleaner shutdown code
- Remove unnecessary realm creation to validate config
- Do not leave timer goroutine running when websocket is closed.
- Do not leave timer goroutine running when rawsocket is closed.

Test code can consolidate cleanup to remove repeated sections of test code that are not really part of any test. Common setup code can add cleanup to `t.Cleanup` to remove the burden of executing cleanup from individual tests.
 
A different package is used for leak detection.

Test code is clarified and reduced by the use of the [`require`](https://pkg.go.dev/github.com/stretchr/testify/require) test package to handle test assertions.

Some small improvements in router shutdown code were made to make it easier to see that all goroutines are ended.

N/A

### What is the new behavior?
N/A

### Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.


KSDaemon:
@gammazero sometimes GH caches original PR description and doesn't take the updated one, so `Lint Commits messages and Pull Request Description / Lint PR & Commits` is failing... Weird....

gammazero:
@KSDaemon If you want to take a look at the changes that are not in `_test.go` code, that would be great. Files are:
- [`router/broker.go`](https://github.com/gammazero/nexus/pull/300/files#diff-5ba094ef20e3c024fa5408bbeeeecf6fbef33b16095157673b84a45cc977e381)
- [`router/dealer.go`](https://github.com/gammazero/nexus/pull/300/files#diff-7aa5dc1d38bb9c17ccc27538fd6e899b936997d99c091c2b37578e5d5e5b0d9a)
- [`router/router.go`](https://github.com/gammazero/nexus/pull/300/files#diff-c399d26e7f23b713ce01946ac9e0e6f1fc94d692e8a086180191901da91b6964)
- [`transport/rawsocketpeer.go`](https://github.com/gammazero/nexus/pull/300/files#diff-1fdd8fd5c51113742470c2de7e67b2f579bc9474a12ec77647fecfd1ab0e8f5e)
- [`transport/serialize/jsonserializer.go`](https://github.com/gammazero/nexus/pull/300/files#diff-9db072a37b04f52f4991392631c95405b39085d43cabdce1288f06fc20f3b115)
- [`transport/websocketpeer.go`](https://github.com/gammazero/nexus/pull/300/files#diff-0fe2d614b29d80ea9015ae6c156fca9a8c3b6f15d973246868b7414c00db4ff8)

KSDaemon:
@gammazero yeah! Sure! Thanks for pointing to files! 

btw, if you're around, can you have a look at my comments in https://github.com/gammazero/nexus/pull/297 ?

aat/auth_test.go
```diff
@@ -67,45 +60,34 @@ func TestJoinRealmWithCRCookieAuth(t *testing.T) {
 
 	if scheme == "ws" || scheme == "wss" {
 		jar, err := cookiejar.New(nil)
-		if err != nil {
-			t.Fatal("failed to create CookieJar:", err)
-		}
+		require.NoError(t, err, "failed to create CookieJar")
```
KSDaemon, line 63: btw, authors suggest that if you assert many times, use the below:
```go
assert := assert.New(t)
// or in our case
assertions := require.New(t)
```

So we can skip `t` first param in every assertion. and also it seems that this should be faster...

KSDaemon (approved):
Ufff... You've done huge test refactoring! 👍🏻 

I like the improvement of waiting for active messages to process before shutdown! 

Ideally, these two things should be split up in different PRs (improvement and tests refactoring), but that is for another time ;)

aat/auth_test.go
gammazero: > authors suggest that if you assert...
As far as using `require.New(t)`, I opted not to do that as all it does is call the corresponding `require.func`. [Example](https://github.com/stretchr/testify/blob/v1.8.4/require/require_forward.go#L132). That ends calling the helper type assertion twice. ...not that I care about performance in test code. Getting rid of that first param did not seem that important, but if you have any strong preference I will be happy to.

gammazero:
@KSDaemon Thank you for the review. I totally agree that fixes/enhancements should be in separate PRs. It was less important in this case since the fixes were mostly to enable better leak detection which was part of the test changes, and did not result in any functional or behavioral changes.

aat/auth_test.go
KSDaemon: Nope It's ok! I think this is a matter of preferences only.

Issue #291, opened 2023-01-05
c.sess.Send error handling


KSDaemon:
Right now all over the sources, we do not check and process the return error value from `c.sess.Send` call.

Golang linter of course is complaining about that. We need to agree what to do:

* Just mark such places as ignored by linter
* Or really check for the error. But then we'll get another question: what to do if there is an error because if that is a really failure — the underlying connection will be closed and there is no need to process the error somehow

KSDaemon:
@gammazero What do you think?

gammazero:
The error response from `Send` is not used, and the error returned by `TrySend` is only used in a couple of places, where it is questionable that it is even necessary.  There really is no need for `Send` to return an error.

I think the best solution is to change `Send` to return a write-only channel. That way the sender can choose how to handle the write. The sender can block, can select with other channels, etc.  This also gets rid of the `TrySend` and `SendCtx` functions and simplifies APIs. Unfortunately, this constitutes a breaking API change and would require a major revision, to follow semver strictly. WDYT?

If we make this change to Send, would it be worth a major revision update?  Should we consider forking the repo to create a new project where we can restart the versioning before 1.0.0 and make many other API and structure changes?

KSDaemon:
I think it is ok. We follow SemVer, right? - Right. So I don't see any problem in sometimes breaking changes if that simplifies the logic and improves the project maintainability and increase the project quality in general. All we have to do: introduce the changes, describe the migration path (v3 → v4) so it will be clear what was changed and how to adopt the new approach and bump the major version.

Regarding forking - I don't think it is needed. We're not a big team just a few enthusiasts so let's keep it convenient for us. We can accumulate all the changes in separate branch, then start publishing the v4.pre, v4.beta and so on while polishing the changes. I don't think we have enough power to support more than one version like v3 and the next one at once. So whenever we release the next major update - we'll probably stop supporting the previous one and will encourage our users to update.

PR #302, opened 2023-09-08
Send API returns a write-only channel

gammazero:
This PR simplifies APIs by removing `TrySend` and `SendCtx`, and allows the caller of `Send` more flexability. More flexibility is provided by a channel as it allows the caller to choose to use it in a select, range loop, etc. when writing a message.

Addresses Issue #291

A message is passed into the blocking `Send` or non-blocking `TrySend`, or a message and context is passed into `SendCtx`.

### What is the new behavior?
The `Send` function returns a write-only message channel. The `TrySend` and `SendCtx` functions disappear.

_This might not be considered a breaking change if `Send` is considered an internal API used when writing your own router or client implementation using the `wamp` package. You can see this is the case since all of the [examples](https://github.com/gammazero/nexus/tree/v3/examples) do not need to change._

### Checklist:

### DIFF
--- a/router/auth/crauth.go
+++ b/router/auth/crauth.go
@@ -89,12 +89,14 @@ func (cr *CRAuthenticator) Authenticate(sid wamp.ID, details wamp.Dict, client w
 	}
 
 	// Challenge response needed.  Send CHALLENGE message to client.
-	err = client.Send(&wamp.Challenge{
+	chalMsg := &wamp.Challenge{
 		AuthMethod: cr.AuthMethod(),
 		Extra:      extra,
-	})
-	if err != nil {
-		return nil, err
+	}
+	select {
+	case client.Send() <- chalMsg:
+	default:
+		return nil, errors.New("cannot send to client: blocked")
 	}
 
 	// Read AUTHENTICATE response from client.
--- a/router/auth/crauth.go
+++ b/router/auth/crauth.go
@@ -96,7 +96,7 @@ func (cr *CRAuthenticator) Authenticate(sid wamp.ID, details wamp.Dict, client w
 	select {
 	case client.Send() <- chalMsg:
 	default:
-		return nil, errors.New("cannot send to client: blocked")
+		return nil, errors.New("cannot send challenge to client: blocked")
 	}
 
 	// Read AUTHENTICATE response from client.
--- a/router/auth/ticket.go
+++ b/router/auth/ticket.go
@@ -74,14 +74,16 @@ func (t *TicketAuthenticator) Authenticate(sid wamp.ID, details wamp.Dict, clien
 		ticket = nil
 	}
 
-	// Challenge Extra map is empty since the ticket challenge only asks for a
-	// ticket (using authmethod) and provides no additional challenge info.
-	err = client.Send(&wamp.Challenge{
+	chalMsg := &wamp.Challenge{
 		AuthMethod: t.AuthMethod(),
 		Extra:      wamp.Dict{},
-	})
-	if err != nil {
-		return nil, err
+	}
+	// Challenge Extra map is empty since the ticket challenge only asks for a
+	// ticket (using authmethod) and provides no additional challenge info.
+	select {
+	case client.Send() <- chalMsg:
+	default:
+		return nil, errors.New("Cannot send chllenge to client: blocked")
 	}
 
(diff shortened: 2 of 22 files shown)


wamp/peer.go
```diff
@@ -1,64 +1,37 @@
...
+	// IsLocal returns true if the session is local.
+	IsLocal() bool
+
 	// Recv returns a channel of messages from the peer.
 	Recv() <-chan Message
 
-	// IsLocal returns true if the session is local.
-	IsLocal() bool
+	// Send returns the peer's outgoing message channel.
+	Send() chan<- Message
```
KSDaemon, line 20: I find naming `Send` confusing. It was clear in previous meaning and value. But looking somewhere in the code at line `peer.Send()` it is not clear that it returns a channel now. And as the signature is changed (thus making new design incompatible with old behaviour) - maybe we can give it better name? At least smth like `SendCh`? What do you think?

router/dealer.go
```diff
@@ -1487,12 +1494,12 @@ func (d *dealer) regCountCallees(msg *wamp.Invocation) wamp.Message {
 	}
 }
 
-func (d *dealer) trySend(sess *wamp.Session, msg wamp.Message) bool {
-	if err := sess.TrySend(msg); err != nil {
-		d.log.Printf("!!! Dropped %s to session %s: %s", msg.MessageType(), sess, err)
-		return false
+func (d *dealer) trySend(sess *wamp.Session, msg wamp.Message) {
```
KSDaemon, line 1497: So we dropped the bool indicator of the result... Was it used somewhere somehow? Do now we miss something because of it?

KSDaemon: Yeah, I saw previously
```
_ = callee.TrySend(&abortMsg)
```

But maybe there are other places which make use of it

router/auth/ticket.go
```diff
@@ -74,14 +74,16 @@ func (t *TicketAuthenticator) Authenticate(sid wamp.ID, details wamp.Dict, clien
...
-	})
-	if err != nil {
-		return nil, err
+	}
+	// Challenge Extra map is empty since the ticket challenge only asks for a
+	// ticket (using authmethod) and provides no additional challenge info.
+	select {
+	case client.Send() <- chalMsg:
+	default:
+		return nil, errors.New("Cannot send chllenge to client: blocked")
```
KSDaemon, line 86:
```suggestion
		return nil, errors.New("Cannot send challenge to client: blocked")
```

router/auth/crauth.go
```diff
@@ -89,12 +89,14 @@ func (cr *CRAuthenticator) Authenticate(sid wamp.ID, details wamp.Dict, client w
...
 		AuthMethod: cr.AuthMethod(),
 		Extra:      extra,
-	})
-	if err != nil {
-		return nil, err
+	}
+	select {
+	case client.Send() <- chalMsg:
+	default:
+		return nil, errors.New("cannot send to client: blocked")
```
KSDaemon, line 99:
```suggestion
		return nil, errors.New("cannot send challenge to client: blocked")
```

KSDaemon: To keep it same in all cases.

wamp/peer.go
gammazero: The reason I left it as `Send` is because that pairs with the existing `Recv` function that also returns a channel. If we change `Send` to something else, then we should also change `Recv`.
```go
// Recv returns a channel of messages from the peer.
Recv() <-chan Message
// Send returns the peer's outgoing message channel.
Send() chan<- Message
```

router/dealer.go
gammazero: Yes, it was used in a few rare places. In those places, a `select` is now used. One such example is [here](https://github.com/gammazero/nexus/pull/302/files#diff-7aa5dc1d38bb9c17ccc27538fd6e899b936997d99c091c2b37578e5d5e5b0d9aR994).

wamp/peer.go
KSDaemon: Yeah, I see. While we're making a refactoring and can give better naming for both, I'm ok to keep it as is. So it's up to you: if you agree - let's rename! If no - no problem.

KSDaemon (approved):
👍🏻 LGTM!

PR #307, opened 2023-11-28
Fix progressive calls: remove options in subsequent calls

KSDaemon:
### Description, Motivation and Context

The WAMP Spec defines that all options except `progress` should be removed from subsequent progressive call messages on the Caller and/or Dealer sides. This PR syncs nexus logic with the spec.

Refs https://github.com/wamp-proto/wamp-proto/pull/511
See [WAMP Spec section](https://wamp-proto.org/wamp_latest_ietf.html#section-11.2-74) for details.

### What kind of change does this PR introduce?

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Checklist:

- [x] My code follows the code style of this project.
- [ ] I have added tests to cover my changes.
- [x] Overall test coverage is not decreased.
- [x] All new and existing tests passed.
- [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly.

PR #308, opened 2023-11-29
RPC forward_timeout option

KSDaemon:

WAMP Spec  in the [Call Timeout section](https://wamp-proto.org/wamp_latest_ietf.html#name-call-timeouts) describes the `forward_timeout` property and how it should be used:

> If the Callee supports Call Timeouts, the Dealer MAY propagate the CALL.Options.timeout|integer option via the INVOCATION message and allow the Callee to handle the timeout logic. If the operation times out, the Callee MUST return an ERROR message with wamp.error.timeout as the reason URI.

> Callees wanting to handle the timeout logic MAY specify this intention via the REGISTER.Options.forward_timeout|boolean option. The Dealer, upon receiving a CALL with the timeout option set, checks if the matching RPC registration had the forward_timeout option set, then accordingly either forwards the timeout value or handles the timeout logic locally without forwarding the timeout value.

This PR introduces such option processing and by the way, fixes the timeout logic. Previously if the timeout was specified by the Caller - it was handled by the router and propagated to the Callee at the same time and only in case when Callee supports it (see snippet below):

```go
	timeout, _ := wamp.AsInt64(invk.options[wamp.OptTimeout])
	if timeout > 0 {
		// Check that callee supports call_timeout.
		if callee.HasFeature(wamp.RoleCallee, wamp.FeatureCallTimeout) {
			details[wamp.OptTimeout] = timeout
		} else {
			timeout = 0
		}
	}
```

This PR depends on https://github.com/gammazero/nexus/pull/307. After that is merged this one will be rebased to reflect only relevant changes.

### Checklist:

### DIFF
--- a/aat/rpc_progress_test.go
+++ b/aat/rpc_progress_test.go
@@ -384,7 +384,7 @@ func TestRPCProgressiveCallTimeout(t *testing.T) {
 	require.Error(t, err, "expected error from CallProgress")
 	var rpce client.RPCError
 	require.ErrorAs(t, err, &rpce, "error should be RPCError type")
-	require.Equal(t, wamp.ErrCanceled, rpce.Err.Error)
+	require.Equal(t, wamp.ErrTimeout, rpce.Err.Error)
 	close(releaseCallee)
 
 	select {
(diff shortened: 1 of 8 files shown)


KSDaemon:
Okay... I'll merge it.

PR #321, opened 2025-02-14
Dealer use session scoped IDs for callee

Skinner927:

Dealers now use [session-scoped IDs](https://wamp-proto.org/wamp_latest_ietf.html#section-2.1.2-11) for `INVOCATION.Request`

Currently, Dealers use Dealer-scoped IDs which means different Callees would receive non-sequential `INVOCATION.Request` IDs.

### What is the new behavior?

IDs for `INVOCATION.Request` are now scoped to the Callee's session.

### Checklist:

### DIFF
--- a/router/dealer.go
+++ b/router/dealer.go
@@ -703,15 +703,12 @@ func (d *dealer) syncCall(caller *wamp.Session, msg *wamp.Call) {
 			request: msg.Request,
 		}
 		d.calls[reqID] = caller
-		invocationID = d.idGen.Next()
 		invk = &invocation{
 			callID:     reqID,
 			callee:     callee,
 			inProgress: isInProgress,
 			options:    msg.Options,
 		}
-		d.invocations[invocationID] = invk
-		d.invocationByCall[reqID] = invocationID
 
 		// Let's check if callee supports this feature
 		// A Callee that supports progressive call invocations, but does not support call canceling,
@@ -812,6 +809,11 @@ func (d *dealer) syncCall(caller *wamp.Session, msg *wamp.Call) {
 			details[wamp.OptProcedure] = msg.Procedure
 		}
 
+		// Only now that we know we're going to send the invocation can we
+		// generate the invocationID
+		invocationID = callee.IdGen.Next()
+		d.invocations[invocationID] = invk
+		d.invocationByCall[reqID] = invocationID
 	} else {
 		// It is an ongoing progressive call (not first one)
 		invk = d.invocations[storedInvocationID]
--- a/wamp/idgen.go
+++ b/wamp/idgen.go
@@ -6,7 +6,7 @@ import (
 	"time"
 )
 
-const maxID int64 = 1 << 53
+const MaxID uint64 = 1 << 53
 
 func init() {
 	rand.Seed(time.Now().UnixNano())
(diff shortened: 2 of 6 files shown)


wamp/session.go
```diff
@@ -161,3 +171,53 @@ func (s *Session) setRoles(details Dict) {
 	}
 	s.roles = roleMap
 }
+
+// UpdateLastRecvID updates the Session's lastRecvID with IDs coming from the
+// other end. Returns true if this is a new ID.
+// This is our only way to track the session-based request IDs to determine if
+// we're responding to a new request.
+func (s *Session) UpdateLastRecvID(id ID) bool {
```
gammazero: Is this called from anywhere?

wamp/idgen.go
```diff
@@ -6,15 +6,15 @@ import (
...
+const MaxID uint64 = 1 << 53
 
 func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
 // NewID generates a random WAMP ID.
 func GlobalID() ID {
-	return ID(rand.Int63n(maxID)) //nolint:gosec
+	return ID(rand.Int63n(int64(MaxID))) //nolint:gosec
```
gammazero, line 17: Need to rebase and resolve conflicts first.

gammazero (changes requested):

wamp/session.go
```diff
@@ -161,3 +171,53 @@ func (s *Session) setRoles(details Dict) {
...
+		return true
+	}
+	// Can only advance at most deltaID
+	// This is specifically to prevent jumping right back into the
+	// rollHighID range after a rollover/wraparound, but't's  it's good genearl
+	if id > last && (id-last) < deltaID {
+		return true
+	}
+	return false
+}
```
gammazero, line 223:
```suggestion
// IsNewRecvID returns true if the ID is considered new.
//
// It is not guaranteed that when lastRecvID is MaxID the next ID will be 1,
// becuase it is possible for a Dealer to skip IDs if there are internal errors
// or if it incorrectly uses Dealer scoped IDs instead of Session scoped. To
// maximize compatability, allow ID rollover/wraparound as long as the
// wraparound distance is withing an allowed limit.
//
// Call this function when the Session.Lock is held.
func (s *Session) IsNewRecvID(id ID) bool {
        const deltaID = ID(500) // maximum allowed wraparound distance

        last := s.lastRecvID

        if id == last {
                return false
        }
        // If the new id is less than the last id, and the wrap-around distance
        // from the last id to the new id is withing delta, then this is a
        // legitimate new id within the allowed wraparound.
        if id < last {
                return (MaxID - (last - id)) < deltaID
        }

        // Can only advance at most deltaID. This is specifically to prevent
        // jumping to close to the last id after a wraparound.
        return id > last && (id-last) < deltaID
}

```

router/dealer.go
```diff
@@ -812,6 +809,11 @@ func (d *dealer) syncCall(caller *wamp.Session, msg *wamp.Call) {
 			details[wamp.OptProcedure] = msg.Procedure
 		}
 
+		// Only now that we know we're going to send the invocation can we
+		// generate the invocationID
+		invocationID = callee.IdGen.Next()
+		d.invocations[invocationID] = invk
+		d.invocationByCall[reqID] = invocationID
```
gammazero, line 816: Using invocationID = callee.IDGen.Next() makes INVOCATION.Request IDs only session-scoped, but d.invocations is still keyed solely by invocationID (dealer-wide). Two different callees will both generate the same invocation IDs (starting at 1), which can overwrite entries in d.invocations and break correlation for YIELD/ERROR/CANCEL/INTERRUPT across concurrent calls. Consider keying invocations by (callee session ID, invocation request ID) (e.g., reuse requestID{session: callee.ID, request: invocationID}) or otherwise ensuring dealer-wide uniqueness for the internal map key while still sending session-scoped IDs to the callee.

Issue #343, opened 2026-03-11
Concurrent map access panic in prepareEvent during websocket send


d-curtis:
**Describe the bug**
Nexus crashes with "fatal error: concurrent map iteration and map write" 
when publishing events while simultaneously sending messages to clients.

**To Reproduce**
Unfortunately don't have a minimal repro just yet, this happens intermittently in a large stack, I can work on trying to isolate this further if needed, but hopefully the log file provides enough of an indication?

**Expected behavior**
deepcopying any shared data / synchronizing read/writes to avoid this panic

**Environment (please complete the following information):**
 - MacOS 15.7.3
 - Nexus Router version 3.3.0
 - WAMP Client - [Autobahn Python](https://github.com/crossbario/autobahn-python) version 24.4.2
 - Heavy Publish traffic with multiple subscribers

**Additional context**
I'm not super familiar with go stack traces to be able to hunt down the root cause, so I might be interpreting this log wrong. My best guess (admittedly with a little help from our AI overlords) is there's a race condition when:
1. Client publishes event with Arguments/ArgumentsKw map
2. Broker's prepareEvent/trySend modifies this map (adding metadata)
3. Simultaneously, websocket transport serializes the same map to JSON
4. No defensive copy is made before either operation

I've attached the stdout/err log from nexusd. 
I noticed something similar in #322 - I had this issue previously with Nexus v3.2.2, but it looks like this update to 3.3.0 hasn't addressed this specific crash.

Just want to say thank you for your hard work on this, I'm currently working on moving away from a different router and so far been super impressed with the performance and portability of Nexus, it's been mostly painless to switch over so far.

Please let me know if there's any more info I can provide or any further troubleshooting to try. Cheers!

[_nexus.log](https://github.com/user-attachments/files/25912234/_nexus.log)

PR #344, opened 2026-03-13
[#343] Copy maps to avoid concurrent access during serialization

d-curtis:

Fixes concurrent access panic described in #343 for websocket clients

`prepareEvent` didn't copy details / args / kwargs for websocket clients

### What is the new behavior?

details / args / kwargs are now copied at the top of `prepareEvent`

- [ ] Enhancement (improve existing code or documentation without affecting behavior)

### Checklist:

---

Worth saying, I did spend some time trying to come up with a stress test to trigger this issue, but couldn't make it fall over on the old implementation, but trying this change with our real app does appear to fix the issue that we were having.
I know it's not satisfying to not have a minimal repro.

I'm also not a go developer, so there may be more elegant ways of doing this, would welcome the feedback.

### DIFF
--- a/router/broker.go
+++ b/router/broker.go
@@ -3,7 +3,6 @@ package router
 import (
 	"fmt"
 	"maps"
-	"slices"
 	"time"
 
 	"github.com/gammazero/deque"
--- a/router/broker.go
+++ b/router/broker.go
@@ -745,16 +745,33 @@ func (b *broker) trySend(sess *wamp.Session, msg wamp.Message) {
 }
 
 func prepareEvent(pub *wamp.Session, msg *wamp.Publish, pubID wamp.ID, sub *subscription, sendTopic, disclose bool, eventDetails wamp.Dict, subscriber *wamp.Session) *wamp.Event { //nolint:lll
-	details := eventDetails
-	if details == nil {
+	// Make defensive copies of all maps to prevent concurrent map iteration and write issues.
+	// These maps may be accessed concurrently by multiple goroutines during serialization.
+	var details wamp.Dict
+	if eventDetails != nil {
+		details = make(map[string]any, len(eventDetails))
+		maps.Copy(details, eventDetails)
+	} else {
 		details = wamp.Dict{}
+	}
+
+	var args wamp.List
+	if msg.Arguments != nil {
+		args = make([]any, len(msg.Arguments))
+		copy(args, msg.Arguments)
+	}
+
+	var argsKw wamp.Dict
+	if msg.ArgumentsKw != nil {
+		argsKw = make(map[string]any, len(msg.ArgumentsKw))
+		maps.Copy(argsKw, msg.ArgumentsKw)
 	}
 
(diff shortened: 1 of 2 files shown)


markope:
Backported this fix locally and wrote a regression test for #343 — opened a PR against the [source branch](https://github.com/d-curtis/nexus/pull/1) so it can fold in here. Test reliably triggers the panic without these defensive copies and goes green with them.

akun-github:
[fix-concurrent-map-race-in-router.patch](https://github.com/user-attachments/files/29281044/fix-concurrent-map-race-in-router.patch)

We had this issue and investigated it independently before seeing this PR. I'm
attaching the patch we came up with.

I also compared each approach's effect on performance and safety, for your
consideration.

**The test from @markope**

We rely on @markope's regression test (d-curtis/nexus#1) rather than writing a
new one: it already exercises exactly the right code path. The test uses
prefix-match subscriptions to force `sendTopic=true`, which triggers the
offending write into `event.Details`. Our patch passes it.

**Scope of copies**

Our patch copies only `Details`. This PR also copies `Arguments` and
`ArgumentsKw` for every subscriber.

The fatal error requires a concurrent writer. After the `Publish` message is
queued, nothing writes to `msg.Arguments` or `msg.ArgumentsKw`. Concurrent
iteration of a map or slice with no concurrent writer is safe in Go, so copying
those fields is not required to fix the crash. @markope's test confirms this:
the large `ArgumentsKw` payload is present to widen the race window on `Details`
(more time spent in the JSON encoder), not because `ArgumentsKw` itself races.

**Performance at high fan-out**

`prepareEvent` is called once per subscriber per publication. This PR copies
`Details`, `Arguments`, and `ArgumentsKw` unconditionally, so every subscriber
pays for three allocations. Our patch copies only `Details` unconditionally.
`Arguments` and `ArgumentsKw` are only copied inside the existing `IsLocal()`
branch, which only fires for the router's own internal meta-event handlers, not
for remote subscribers.

In a deployment with thousands of remote subscribers, the `IsLocal()` branch is
essentially never taken. Our patch costs one allocation per subscriber per
publish. This PR costs three.

**The existing `IsLocal()` block**

This PR removes the local-subscriber deep copy block, since all three fields are
now copied unconditionally. Our patch trims the block instead: the `Details`
copy inside it is now redundant (Details is already a private map by the time
`IsLocal()` runs), so we remove it. The `Arguments` and `ArgumentsKw` copies
stay, since those still protect against a local handler mutating the original
message.

**dealer.go**

This PR copies `Arguments` and `ArgumentsKw` in `syncCall` and `syncYield` but
not in `syncError`, which follows the same pattern. Our patch covers all three,
and adds the same defensive copies with the same rationale: nothing currently
writes to those fields after the message is queued, but aliasing a session's
incoming message data into an outbound message for a different session's
transport goroutine is the same structural pattern that causes the fatal broker
race. The cost is one allocation per RPC transaction, constant regardless of
subscriber count, and negligible compared to the network round-trip.

